PR #1, opened 2023-02-09
Saavik

saavikford:
This is a reasonably complete input setup, and could probably be implemented in main with a bit of careful reading; however, there are things that *used* to be under 'input' that are really housekeeping things, that should be put somewhere else, so if you just delete the old input section, and replace it with 'ReadInputs.py' you may end up very sad. I have kept track of those extra variables in the original McFACTS repo... but not here. Also note, I typecast numbers that are counting (so number of time steps, number of bh) to int, rather than float, and changed them in the 'model_choice.txt' file as well.


saavikford:
I have only been mucking in inputs. You can ignore it or use it... should not create a problem to merge (and merging will make it possible to use it in main).

PR #15, opened 2023-11-14
Add work directory capability

cookhe:
- New command line argument `-w, --work-directory` sets an user-defined output destination.
- Now the script can be run using `/path/to/test1.py` while located in `/some/other/path` and will properly find files designated by relative paths to `test1.py` (i.e., `inputs/*`).
- The working directory is set to the current location by default.
- Raises a `FileNotFoundError` exception if user-defined work directory does not exist.


saavikford:
I think this is only adding the optional work directory from the command line so I'm ok'ing it!

PR #39, opened 2024-04-09
Xevra

xevra:
- Added vim swap files to gitignore
- futureproofed instance of np.int for python 3.11 in setupdiskblackholdes.py
- Attempted to patch test1_iterate.py to match behavior of test1.py (previously wasn't running at all). There may still be some dynamics missing.


bmckernan (approved):
Yes

PR #42, opened 2024-04-26
Reorganized for pip installation

xevra:
Alright, this is a doozy.

1. The package is now pip installable from source (not distributed).
  a. **Most of the McFACTS code is now located in /src/mcfacts/, and divided into submodules.
      It's the same code that we know and love, but organized more like other python packages.**
  b. There is a new setup.py file which pip uses to know how to install the package.
      It has all of the requirements (numpy, scipy, matplotlib, and now pandas) listed as dependencies, so pip will install them automatically when installing mcfacts.
      We should go through this together and discuss licensing and metadata for the package before we make this public.
  c. Some of the data files are accessed slightly differently, through python's import resources.
  d. `__version__.py` files are generated by the Makefile I added. They all read 0.0.0. We should not change this unless we are ready to make a detailed changelog when we change things, or have another good reason to.
  e. I renamed all the `__init.py__` files to `__init__.py`. The former may work, but they bothered me a lot.

2. Scripts have their own directory, in /scripts
  a. After discussing it with Harry, test1.py was not really what we expected a 'test' to be. In the future, we may want a tests folder, but those tests are meant to test individual functions, not the entire module.
      For that reason, test1.py has been renamed to scripts/mcfacts_sim.py
  b. The test1_iterate has been removed as it was deprecated.
  c. The plotting script has been moved to /scripts/population_plots.py
  d. All scripts use the library to import packages, rather than relative filepaths, allowing a similar script to be run from anywhere.

3. model_choice.txt has been renamed to /recipes/model_choice.ini
  a. The change of the file extension helps some text editors / IDEs to display the standard .ini colors. 
      We were already calling it an 'inifile', so I felt that was okay.
  b. The recipes folder helps keep the directory structure cleaned up.
  c. I added a line to the top of the ini file, so that the parser would have an easier time with it.

4. Some duplicate files from the vis directory (now /src/mcfacts/vis) were removed from the top level directory. I checked that these files were exactly the same using the `diff` command before deleting anything.

5. Changes to /src/mcfacts/inputs/ReadInputs.py
  a. After discussing it with Harry, we both wanted to see this file return a dictionary rather than several lines of variables names. Harry pointed out that some arguments were not even passed to return, and some were passed multiple times.
  b. Because of changes to the inifile, we don't need to artificially add a `top` line anymore.
  c. Rather than potentially changing data types more than once, I created a dictionary of data types for the identified arguments. For arguments *not* identified in the dictionary, they will use the old scheme for determining their type.

6. Changes to /scripts/mcfacts_sim.py
  a. Imports now draw from the mcfacts library rather than relative pathing.
  b. The argument parser now has its own special function.
  c. The argument `--use-ini` has been replaced by `--fname-ini`, and the default has been updated.
  d. The argument parser now accepts mcfacts inputs as arguments (i.e. `--mass_smbh`).
      This works even though they are not all explicitly typed out, and will work for anything identified in the default initialization file.
      You also don't need to pass a 'complete' ini file. You only need to pass the arguments you want to change.
      The priority is command line arguments > user defined `fname_ini` > default `fname_ini`
  e. To avoid polluting the namespace, the arguments are stored in opts throughout the script.
      I did my best to make this change everywhere appropriate, and tested the script multiple times, but I might have missed something.
      If you run a test, and find a variable is missing, try `opts.[variable-name]`
  f. I made a lot of really long lines multiple lines.
  g. Added shebang to first line.

7. Changes to /scripts/population_plots.py
  a. Scooped up code into main function.
  b. Created an argument parser.
  c. library imports replace relative path imports.
  d. O3 data files are accessed through import resources.
  e. Added shebang to first line.
  
8. Makefile -- I made this because I wanted to
  a. Changing the version in the makefile and running the command `make version` will update the `__version__.py` files.
  b. `make install` will run `pip3 install -e `; This will install the package in such a way that you can still modify python files without reinstalling the package for each change.
  c. Just running `make` will run the test script (/scripts/mcfacts_sim.py) and plotting script (/scripts/population_plots.py).
      This is how I test that the code works before I submit any changes.
  d. Running `make clean` from the command line will remove all of the junk created during the test outlined above.

9. I updated the README to instruct users that they can now use pip to install the package from source.


xevra:
Quickly added an option to dump input arguments into log file.

xevra:
Fixed a bug with __version__.py

cookhe (commented):
Everything looks good to me when we ran it on meeting last week.

PR #62, opened 2024-05-28
Added code to write file with initial BH parameters before entering the time loop

spacekaila:
Code now writes a file named `initial_params_bh.dat` in each run directory. File has header `R M a theta orb_ang_mom orb_ecc orb_incl`, corresponding with location, mass, spin, spin angle, orbital angular momentum, orbital eccentricity, and orbital inclination.


spacekaila:
Superseded by #79

PR #101, opened 2024-08-21
change to pyproject.toml, update Makefile with new install command

JakePost:


xevra (approved):
Looks great!

I am ecstatic to see better ways of doing things.

A few notes (not problems):
- Previously, the default installation method for `make install` was in editable mode. I agree that should not be the standard for users who are not developers. However, I think we do need to communicate to the team that the behavior of `make install` has changed.
- Authors list should include Kaila and others as well, but it didn't before, so it shouldn't hold up the merge request.

Thank you!

JakePost:
Long term I think its ok for it to be editable. Especially if we decide to target Jupyter / iPython / third-party use in the future. People won't have to worry about running the make file and will just be running from their environment cache.

PR #105, opened 2024-08-28
Main dev

saavikford:
trying to update my version to main-dev while preserving all the retro upgrades I implemented... this is probably going to make someone cry... probably me.


saavikford:
I'm merging main-dev into my branch which also contains retro upgrades... then I'll run it, and if everything is ok, I will merge back to main-dev so everyone can enjoy the joy of retro orbiters...

PR #111, opened 2024-09-09
Single black holes are now using the AGNObject class, global random number generator, a new filing cabinet

spacekaila:
**Black holes**
- Prograde and retrograde single black holes are fully using the AGNObject class.
- The existing prograde_bh_[param] and retrograde_bh_[param] arrays have all been commented out.
- All functions use the parameters from the AGNBlackHole object, instead of the old arrays.
- The binaries, EMRIs, and inner circle BHs are unchanged.
- Black holes now also have a unique ID number --this will be in use later when other types of objects are interacting in the disk.
- AGNBlackHole objects can add new BHs, remove BHs, create a new AGNBlackHole object with only BHs at specified indices, copy the AGNBlackHole object, return a list of parameters, dump to file, and initialize from file (currently untested).

**Random number generator**
The random number generator is now global, so you only need to put `from mcfacts.mcfacts_random_state import rng` and then can use rng as normal, without having to include it in function calls. This is fully implemented in mcfacts_sim.py, but only the setupdiskstars.py and setupdiskbh.py have been modified to no longer require rng in the function inputs.

**Other things**
-There is partial implementation of AGNFilingCabinet, which is intended to be a master catalog of all objects in the disk. The filing cabinet has a unique ID number for each object, which can then be used to locate it in its respective AGNObject. The idea is that we can scan through everything in the disk using the filing cabinet to determine which objects will interact, then pull those objects via their ID numbers and make them do things.


saavikford:
Discussed during Hack Week--discussion recorded via zoom. NOTE: data structures have CHANGED!!

PR #121, opened 2024-09-10
Standardize variable names and add doc strings into main-dev

cookhe:
Linted all files. Updated the variable names to match the new convention. Added doc strings to functions.

Removed unused function: `bin_tgw()` from `tgw.py`


cookhe (commented):
Looks good!

PR #196, opened 2024-09-30
Xevra bugfix 240929

xevra:
Three sets of changes:

Bugfix 1:
A previously implemented check for unphysical binaries was not correctly removing them from the AGNFillingCabinet.
Additionally, if there were any unphysical binaries, all of the binaries were frozen for a given time step.

Bugfix 2:
The new version of the baruteau function was causing any binary with eccentricity greater than one to merge.
The old version of the baruteau function was preventing any binary with eccentricity greater than one to merge.
The solution is a new unphysical check, which marks any binaries with an eccentricity greater than, or equal to 1 for deletion.

New:
I added my new plotting function for the pAGN models.
These won't work unless you comment the currently broken truncation code in DiskModelsPAGN.py
The reason I have refrained from doing so is to limit merge conflicts with Harry's branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@ mergers_nal*.png
 r_chi_p.png
 *nohup.out
 run*
+*disk_interp_rho.png
+*pAGN_*_disk_rho.png
+*pAGN_*_disk_tauV.png
+*pAGN_*_disk_T.png
 
 # Operating System Files
 .DS_Store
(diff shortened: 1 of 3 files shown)


JakePost:
Graphs are consistent when compared to Barry's baseline & current main-dev. 

100 Galaxies, pAGN Disabled
Run arguments: [out.log](https://github.com/user-attachments/files/17197141/out.log)

q_chi_eff
r_chi_p
time_of_merger
merger_remnant_mass
merger_mass_v_radius
m1m2
gw_strain

JakePost (approved):

PR #197, opened 2024-09-30
Update random number generator usage for proper reproducibility

cookhe:
The McFACTS random number generator now uses the `np.random.RandomState()` object class.
Documentation Link: [RandomState v.1.26](https://numpy.org/doc/1.26/reference/random/legacy.html)

1. change calls from `rng.random()` to `rng.uniform()` since the former is not a method in the documentation despite functioning if used.
2. remove superfluous type checking `int(int)` 
3. remove deprecated and unused function `setup_disk_blackholes_inclination()` since it has been superseded by `setup_disk_blackholes_incl()`
4. change calls from `rng.integers()` to `rng.randint()`

### DIFF
--- a/src/mcfacts/mcfacts_random_state.py
+++ b/src/mcfacts/mcfacts_random_state.py
@@ -3,7 +3,7 @@ import numpy as np
 
 default_seed = 1
 # Initialize random number generator
-rng = np.random.default_rng(default_seed)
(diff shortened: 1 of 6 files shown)


xevra:
I'm still seeing 1 in 100 galaxies are slightly different.

I'm going to merge anyway because I think our problems may be caused by Undefined Behavior rather than a broken random number generator.

PR #249, opened 2024-11-11
Add terminal text after setup command, fix python version requirement

JakePost:


Makefile
```diff
@@ -53,9 +53,15 @@ setup: clean version
 	source ~/.bash_profile && \
```
cookhe, line 53: Running `$ make setup` fails here because I do not have a `~/.bash_profile` file. Looking in the commit record, @JakePost updated this in commit [7453fda](https://github.com/McFACTS/McFACTS/commit/7453fdac97cefb5bd9ecdb151fa8b2ce26951622) to prepare for CLI. Is this make target not intended to be ran by a human?

spacekaila: The default shell on Mac is now zsh, but some people still use bash. Is there a way to figure out what user's shell is and source the right file?

cookhe: Should be able to use the `$SHELL` environment variable.

spacekaila: I'm also a bit leery of changing the user's shell profile. Most other codes I've used just have lines to set whatever variables you need for that session and have instructions on how to alter your shell profile if you want those variables set permanently.

PR #250, opened 2024-11-11
Fixing bugs in type1_migration and retro_bh_orb_disk_evolve

spacekaila:
If an object tried to migrate out of the disk, we set its `orb_a  = disk_outer_radius`, which caused problems later on if two BH at `disk_outer_radius` formed a binary (`orb_a_1 = orb_a_2` so `bin_sep = orb_a_1 - orb_a_2 = 0`). Now a BH migrating out of the disk has its `orb_a = disk_outer_radius - epsilon` where epsilon is the Hill sphere of the BH at `disk_outer_radius` times a random number between 0 and 1.

When evolving retros, objects that had their `orb_a <= 0` in the calculations were caught and were set to `orb_a = 0.0` This has been changed so that they are set to `disk_inner_stable_circ_orb` instead.

### DIFF
--- a/scripts/mcfacts_sim.py
+++ b/scripts/mcfacts_sim.py
@@ -1220,7 +1220,6 @@ def main():
                     agn_redshift
                 )
 
-
             if blackholes_inner_disk.num > 0:
                 blackholes_emris.add_blackholes(new_mass=blackholes_inner_disk.mass,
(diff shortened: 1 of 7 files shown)


spacekaila:
Also fixed masking bug

cookhe (approved):
Checked run and plots match @spacekaila's

PR #251, opened 2024-11-14
fixed bug in excluded_angles in bin_spheroid_encounter

spacekaila:


cookhe (approved):
Ran the default model. Plots look reasonable.

PR #256, opened 2024-11-26
Fixed bug in eccentricity functions, red scare is real. More catches for orb_a > disk_radius_outer.

spacekaila:
Fixed a bug in `dynamics.circular_singles_encounters_prograde` where circular BHs were defined as `orb_ecc <= ecc_crit`, but eccentric BHs were defined as `orb_ecc >= ecc_crit`, leading to overlap. Eccentric BHs are now defined as `orb_ecc > ecc_crit`. This created the red scare.

Also fixed a similar bug in `eccentricity.orbital_ecc_damping` and `eccentricity.bin_ecc_damping` where `modest_ecc_prograde_indices` were defined as `orb_ecc <= 2 * disk_aspect_ratio` and `large_ecc_prograde_indices` were defined as `orb_ecc >= 2 * disk_aspect_ratio`. We now define `large_ecc_prograde_indices as orb_ecc > 2 * disk_aspect_ratio`.

If a binary is ionized (broken apart) the two new single BHs have their `orb_ecc` set as a random number between `0` and `orb_ecc_max` (set in ini file). This function is `eccentricity.ionized_orb_ecc`.

If a binary merges, the new single BH has its `orb_ecc` set as `v_kick/v_kepler`, with `v_kick` currently hardcoded to 200 km/s for all BHs. This function is `merge.merged_orb_ecc`.

If a binary forms from two single BHs, it's bin_orb_ecc is set to ecc_crit (set in ini file). This function is `formation.add_to_binary_obj`.

Added catches for objects trying to move out of the disk in `dynamics.circular_binaries_encounters_ecc_prograde` and `dynamics.circular_binaries_encounters_circ_prograde`. Added catch for objects set to the `disk_radius_trap` in `migration.type1_migration`. In all cases, object's `orb_a` is set to `disk_radius_outer - epsilon` or `disk_radius_trap - epsilon`, where `epsilon` is defined as the Hill radius at `disk_radius_outer` or `disk_radius_trap` multiplied by a random number between 0 and 1. This ensures that two objects do not have exactly the same `orb_a`, preventing them from forming a binary with a separation of zero.

### DIFF
--- a/misc/retro_ecc.py
+++ b/misc/retro_ecc.py
@@ -1,7 +1,7 @@
 """ The module provides methods for calculating the eccentricity and semi major latus of retrograde orbiters."""
 import numpy as np
 import scipy
-import mcfacts.constants as mc_const
+import astropy.units as u
 
 
 def retro_semi_lat(smbh_mass, disk_bh_retro_orbs_a, disk_bh_retro_masses, disk_bh_retro_orbs_ecc,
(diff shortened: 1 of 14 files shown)


bmckernan (approved):
I agree that the red scare seems real. Approve this request.

PR #257, opened 2024-11-27
Stars are now functional in McFACTS!

spacekaila:
We have stars!!!!!!!

They don’t interact with anything, but they accrete mass (at an alarmingly high rate) and migrate in the disk!!

- Values for log Teff and log L are interpolated off https://ui.adsabs.harvard.edu/abs/2022MNRAS.511.2814Y/abstract and right now stars are assumed to always be at ZAMS because they accrete so fast.
- Stars are created with an IMF between 0.5 and 10Msun and initially run through a Hill sphere coagulation code where stars that are in each others’ Hill spheres are coagulated together.
- As stars move through the code they accrete mass, but their max mass is set to the immortal cutoff, so if a star tries to accrete enough mass that it would be above the immortal cutoff, it’s mass is just set to the immortal cutoff.
- At the end of a run, any stars that are at the immortal mass are assumed to explode and turn into BHs (and are added to blackholes_pro).

There are a couple star options now.

- `flag_add_stars = 1` if you want stars to exist, `0` if not
- `flag_initial_stars_BH_immortal = 1` if you want the stellar initialization (aka this happens when stars and BHs are created and does not happen again in the time loop) to turn any stars above the immortal cutoff into BHs. If you set it to `0` then stellar initialization just sets stars with mass > immortal cutoff to have mass = immortal cutoff.
- `disk_star_initial_mass_cutoff = 300` is the immortal cutoff
- `disk_star_scale_factor = 1e-3` is a scale factor for the number of stars. Number of stars is set to `num_bh / nsc_ratio_bh_num_star_num * disk_star_scale_factor`.

An average mcfacts run with 201 BHs and 201 stars now produces 2 binaries, 55 mergers, and 160 immortal stars.


JakePost:
On hold for versioning!

bmckernan (dismissed):

bmckernan (dismissed):

bmckernan (approved):
Agreed

PR #260, opened 2024-12-10
Bug fixes in bin_contact_check, bin_harden_baruteau

spacekaila:
#### Summary

We were comparing `time_to_merger_gw` (in seconds) to `timestep_duration_yr` (in years) in `bin_harden_baruteau`. This is now fixed.

Updated the calculation for `time_to_merger_gw` to use the `point_masses.time_of_orbital_shrinkage` function, with `sep_final` set to `R_sch_1 + R_sch_2` with `R_sch` as the Schwarzschild radius.

Updated the calculation in `bin_contact_check` to use the actual Schwarzschild radius of each BH.

Added `r_schwarzschild_of_m` function in `point_masses.py` to calculate the Schwarzschild radius ($R_\text{sch} = 2G M/c^2$).

### DIFF
--- a/src/mcfacts/objects/agnobject.py
+++ b/src/mcfacts/objects/agnobject.py
@@ -28,7 +28,7 @@ attr_binary_bh = ["id_num", "orb_a_1", "orb_a_2", "mass_1", "mass_2", #"mass_tot
 attr_merged_bh = ["id_num", "galaxy", "bin_orb_a", "mass_final",
                   "spin_final", "spin_angle_final",
                   "mass_1", "mass_2",
-                  "spin_1", "spin_2",
+                  "spin_1", "spin_2",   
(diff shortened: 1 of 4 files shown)


src/mcfacts/objects/agnobject.py
```diff
@@ -28,7 +28,7 @@
 attr_merged_bh = ["id_num", "galaxy", "bin_orb_a", "mass_final",
                   "spin_final", "spin_angle_final",
                   "mass_1", "mass_2",
-                  "spin_1", "spin_2",
+                  "spin_1", "spin_2",
```
xevra: (trivial) linting oopsie

spacekaila: Fixed!

bmckernan (approved):
Agreed!

PR #262, opened 2025-01-07
dynamics updated, default run set to 0.5 Myr

spacekaila:
#### Summary

Dynamics functions for binaries now return the altered single object orb_an and orb_ecc. Default run is set to 0.5Myr.


bmckernan (approved):
Looks good.

PR #271, opened 2025-02-28
Big stars addition

spacekaila:
#### Summary

- Stars lose mass via stellar winds
- Merged and exploded stars are tracked in their own objects and written to data files
- Amount of mass added to disk/lost from disk due to stars is tracked and written to data file
- Stars interact with each other in dynamics
- Stars interact with single BHs in dynamics
- If two stars are within each other's Hill spheres, they merge
- If a BH comes within a star's HS, the star explodes and the BH accretes at Eddington rate for one tilmestep
- Function for close encounters checks for stars and BHs (and follows above strategy)
- I think that's all??


JakePost:
Does this PR add some more stuff to the return of the disk model? If so that seems to be why the test is failing again.

spacekaila:
yes! It adds disk_pressure_grad, disk_omega to the return of the disk model

xevra:
A fresh checkout of the stars branch currently has a broken astropy_const call. Looking into it.

xevra:
Fixed some broken astropy calls in a few different places, and pushed to stars. I hope that's okay.

xevra:
At a glance, the plots from `make plots` don't look any different and therefore pass.

Reviewing new content and line by line changes now.

JakePost (dismissed):

xevra:
Line by line looks good for the most part. Some of it is physics that I don't know (specifically, in dynamics, star interactions, and stellar interpolations).

I also need the right inifile to test the movie features.

xevra:
Movies work

xevra (approved):

PR #272, opened 2025-03-11
Added data files for the new disk parameters and fixed bug in pressure grad interpolator

spacekaila:
#### Summary

- Fixed bug in pressure grad interpolator (mistake in implementation of np.gradient)
- Added data files for pressure gradient, omega, temperature for use when pAGN is off


xevra:
mcfacts_sim.py runs successfully when mcfacts is installed with the `--editable` option.

The only issue I have with the files is that `generate_disk_param_files_from_pagn.py` should live in the scripts folder.

However, there are two places where we need to keep track of the datafiles which are passed around and read through impresources.

Those places are the MANIFEST.in file and setup.py.

To really test that this is working properly, you need to run `make test-build`. Right now, this fails.

xevra:
Looks good. I think the files should also be listed in MANIFEST.in, but I forget why. For the purposes of Pypi, it may be enough to put them in setup.py.

xevra:
The test runs, so I'm going to approve it, but I recommend adding them to the MANIFEST.in file before merging.

xevra (approved):

PR #274, opened 2025-03-13
populate disk with BH according to power law probability

cookhe:
#### Summary
1. Black hole locations are now drawn from a volume-scaled broken power law probability distribution. This occurs during initial setup and during the black hole orbit capture process. The old function that used a uniform probability distribution is still available. The user just needs to call it instead if so desired. 
2. A test for this function now exists in `tests/setup/` and will be automatically ran during the test workflow, and I added relevant parameters to the dictionaries in `conftest.INPUT_PARAMETERS`.
3. Fixed a logic error in `dynamics.py` that excluded the equals-to case.
4. Objects that would migrate outward past `disk_radius_outer` are now placed at `disk_radius_outer`
5. Added some comments where I thought they would be useful and trimmed some trailing white space.


JakePost (approved):
100 Galaxies, 0.5 Myr

gw_strain (1)
m1m2
merger_mass_v_radius
merger_remnant_mass
q_chi_eff
r_chi_p
time_of_merger

PR #275, opened 2025-03-16
fixed bug that always initialized BHs and stars with galaxy = 0

spacekaila:
#### Summary

When initializing stars and BHs at the beginning of the time loop, `galaxy` was set with np.zeros, instead of set to the actual galaxy (iteration) we were on. This is now fixed.


cookhe (commented):
Please update the doc string for in the `AGNObject` class definition to state the dimensions of the arrays should match the number of objects in the class. Alternatively, you could add this to the `__init__()` doc string.

spacekaila:
Updated `AGNObject` docstring to say that array length must match number of objects.

cookhe (approved):
Comparison runtime: 507.20 s.
Plots: ✅

PR #276, opened 2025-03-18
Adding boolean switch to go between velocity models. Transfered velocity plots from emily_plots to population_plots

emilymcpike:


JakePost (approved):
looks good

PR #278, opened 2025-03-26
Read log file capability

cookhe:
#### Summary
I made a new module to read output files. For now, it contains one function to read the McFACTS log file and is currently used in `/scripts/population_plots.py`.

##### Details
1. Create a new module file `src/mcfacts/outputs/ReadOutputs.py` containing the `ReadLog` function.
2. Log file name is now consistent across McFACTS. Previously we had cases of `out.log` and `mcfacts_sim.log`. I implemented `mcfacts.log` as the standard. Users still have the option to customize this.
3. The `scripts/population_plots.py` plotting routine uses the new log dictionary, and all radial distributions now include a vertical line at the trap radius. 

##### Possible Tests for Reviewer
The `ReadLog` function generates a dictionary using pre-defined types since the log file contains multiple types. If the log file contains extra keys that are not known to the code, they will still be added but as a string type with a warning communicating this to the user. Test this by adding these lines to a default McFACTS run log file, importing the function into a python environment, and passing the file to the function.
```
flag_bh_are_cool: True
love_factor: 1.4
num_aliens: 10000
```

### DIFF
--- a/src/mcfacts/outputs/ReadOutputs.py
+++ b/src/mcfacts/outputs/ReadOutputs.py
@@ -1,0 +1,285 @@
+"""Define output handling functions for McFACTS
+
+Logfile
+-------
+    "bin_num_max"                   : int
+        Maximum allowable number of binaries at a given time. Sets array size.
+    "fname_ini"                     : str
+        Name of the ini file used for this run.
+    "fname_output_mergers"          : str
+        Name of the output file for mergers.
+    "fname_output"                  : str
+        Name of the output file for the run.
+    "fname_snapshots_bh"            : str
+        Name of the output file recording BH info at each snapshot.
+    "saves_snapshots"               : bool
+        True: save snapshots, False: don't save snapshots
+    "verbose"                       : bool
+        True: print extra info, False: don't print extra info
+    "work_directory"                : str
+        The working directory used for the run.
+    "seed"                          : int
+        Random seed used for the run.
+    "fname_log"                     : str
+        Name of the log file storing details of the run.
+    "runtime_directory"             : str
+        The runtime directory used for the run.
+    "disk_model_name"               : str
+        'sirko_goodman' or 'thompson_etal'
+    "flag_use_pagn"                 : bool
+        Use pAGN to generate disk model?
+    "flag_add_stars"                : bool
+        Add stars to the disk
+    "flag_initial_stars_BH_immortal": float
+        If stars over disk_star_initial_mass_cutoff turn into BH (0) or hold at cutoff (1, immortal)
+    "smbh_mass"                     : float
+        Mass of the supermassive black hole (solMass)
+    "disk_radius_trap"              : float
+        Radius of migration trap in gravitational radii (r_g = G*`smbh_mass`/c^2)
+        Should be set to zero if disk model has no trap
+    "disk_radius_outer"             : float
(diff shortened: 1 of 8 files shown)


src/mcfacts/outputs/ReadOutputs.py
xevra, line 1: It would be better practice to start by importing the INPUT_TYPES dictionary.

Having duplicate INPUT_TYPES and OUTPUT_TYPES dictionaries creates the possibility for INPUT_TYPES[A] to be different from OUTPUT_TYPES[A], which would create confusion about the intended data type for a particular variable.

I understand that there are entries in the logfile outputs which are not present in the inifiles, and so having a separate OUTPUT_TYPES dictionary is a good idea.

I also understand that there may be variables in the logfile with the same name as variables in the inifile, which have different data types.

However, I would address these problems by initializing OUTPUT_TYPES as an empty dictionary, and add entries to it from INPUT_TYPES in a loop, after importing INPUT_TYPES. Then adding new or modified entries to the OUTPUT_TYPES dictionary after that loop.

I understand this is a bit pedantic, so if you don't feel like doing this, we can bring it up on Monday with the group and decide collectively that we don't care.

src/mcfacts/outputs/ReadOutputs.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Loop through the lines
+    for line in lines:
+        # Split the line
+        key, value = line.split('=')
+        # Strip the values of any whitespace
+        key = key.strip()
+        value = value.strip()
+        # Convert the values
+        if key in OUTPUT_TYPES:
```
xevra, line 86: We need a special case for boolans (E.g. `if isinstance(OUTPUT_TYPES[key], bool): log_dict[key] = bool(int(value))`).

xevra: This is because bool('0') -> True

xevra (changes requested):
I would like to see the changes requested in previous comments discussed or addressed before we merge the PR.

src/mcfacts/outputs/ReadOutputs.py
xevra: My example code is wrong. `isinstance(bool,bool)` fails. This should be closer.

```
if isinstance(OUTPUT_TYPES[key](value), bool):
    log_dict[key] = bool(int(value))
```

src/mcfacts/outputs/ReadOutputs.py
cookhe: I agree. We should avoid duplication where possible. I intended to write a developer/user guide that would explain where new input/output parameters must be added to docstrings/dictionaries, etc. Using the existing `INPUT_TYPES` dictionary will simplify that process and reduce the circumstances under which a user triggers the warning.

For now, the log file types for objects listed in the `.ini` file remain the same, so your solution is an easy fix.

cookhe:
Additional changes:

1. Standardized types for flag parameters to be integers across documentation
2. Before writing to the log file, convert any booleans that are introduced to `opts` throughout the code to integers.
3. The `OUTPUT_TYPES` dictionary appends the `INPUT_TYPES` dictionary rather than duplicating it locally.

xevra (approved):
Looks good. Make plots runs. :+1:

PR #281, opened 2025-04-14
Fixed initial star number code

spacekaila:
#### Summary

Added flag to options: `opts.flag_coalesce_initial_stars`. If set to 0, then the initial number of stars is taken from `setupdiskstars.setup_disk_stars_num` and parameters are set accordingly. If set to 1, then the initial number of stars is taken from `setupdiskstars.setup_disk_stars_num`, then stars within mutual Hill spheres are merged together before the beginning of the time loop. This is only necessary if you are starting with a large number of stars (>> 1000). By default the flag is set to 0 and the number of stars is set to ~1000.

### DIFF
--- a/IOdocumentation.txt
+++ b/IOdocumentation.txt
@@ -191,18 +191,44 @@ star cluster variables second, and assorted other choices (time resolution etc.)
         Saving the snapshots of each object being per time step. Deafult is 0 [off] 
         default : 0
 
-    # =========================== DONT TOUCH STARS YET SINCE NOT ADDED ===========================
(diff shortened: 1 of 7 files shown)


cookhe:
Confirmed that `flag_coalesce_initial_stars` set to 1 (on) produces less stars than when 0 (off) because they're treated collectively.

cookhe (approved):
Confirmed that flag_coalesce_initial_stars set to 1 (on) produces less stars than when 0 (off) because they're treated collectively.

PR #289, opened 2025-04-28
Moved torque interpolators into their own function, fixed misspelling

spacekaila:
#### Summary

- Moved interpolators for d log(Sigma) / d log(R), d log(temp) / d log(R), and d log(midplane pressure) / d log(R) into their own function that is called at the beginning of mcfacts_sim
- Deleted `paardekooper10_torque_binary` function as binaries can just use the `paardekooper10_torque` function
- Changed jim*I*nez_masset to jim*E*nez_masset


xevra (approved):
Looks good, and works.

PR #294, opened 2025-04-30
Fix analytical velocity math, fix kick velocity histogram, delete unused luminosity-related mod mock_phot.py, rename emily_plots -> em_plots, commenting through analytical velocity mod.

emilymcpike:


src/mcfacts/objects/agnobject.py
```diff
@@ -1319,9 +1319,8 @@ def __init__(self,
                  chi_eff=empty_arr,
                  chi_p=empty_arr,
                  v_kick=empty_arr,
-                 lum_shock=empty_arr, # emily add
```
spacekaila, line 1322: If you add a parameter to any of the AGNObject subclasses or the AGNFilingCabinet, you need to also add those parameters to the relevant `attr_{THING}` list at the top of `agnobject.py`. Please fix!

src/mcfacts/physics/analytical_velo.py
```diff
@@ -1,68 +1,94 @@
...
 
-#     # Generate random numbers from gamma distribution
-#     r = gamma.rvs(a=shape, scale=scale, size=length)
-#     return r
+    Returns
+    -------
+    v_kick : np.ndarray
+        Kick velocity [km/s] of the remnant BH with :obj:`float` type
+    """
+    mass_1 = np.array(mass_1)
```
spacekaila: You can assume that these things will all be np arrays already

src/mcfacts/physics/analytical_velo.py
```diff
@@ -1,68 +1,94 @@
...
+def analytical_kick_velocity(
+        mass_1,
+        mass_2,
+        spin_1,
+        spin_2,
+        spin_angle_1,
+        spin_angle_2):
+    """
+    Compute the analytical gravitational wave recoil (kick) velocity for merging black hole binaries
+    as in Akiba et al. 2024 (arXiv:2410.19881).
```
spacekaila, line 19: Please write out the details of the model you're using, e.g., section of paper, equation numbers, etc. We want to make it easy for a random user to cross-check a function with the paper it's based on.

src/mcfacts/physics/analytical_velo.py
```diff
@@ -1,68 +1,94 @@
+"""
+Module for calculating the recoil (kick) velocity of a remnant BH.
+"""
 import numpy as np
 from mcfacts.mcfacts_random_state import rng
-from astropy import units as u
-from astropy import constants as const
 
+def analytical_kick_velocity(
```
spacekaila, line 9: Could you add (astropy) units to this function?

src/mcfacts/physics/lum.py
```diff
@@ -3,58 +3,72 @@
...
+    -------
+    Lshock : float
+        Estimated shock luminosity (in erg/s).
     """
-
     r_hill_rg = bin_orb_a * ((mass_final / smbh_mass) / 3)**(1/3) 
     r_hill_m = si_from_r_g(smbh_mass, r_hill_rg)
 
-    r_hill_m = r_hill_m.to('cm').value
+    r_hill_cm = r_hill_m.cgs.value
```
spacekaila: If you don't use `.value` the units will follow through to the end and you can use that as a double check that your final value is in the right units and make it easy if you want it in different units. You can use `.cgs` or `.to("meter")` (or whatever else you want, it'll also accept combinations like "kg/m^2"). You just have to put something like `L_shock = (E/time).to("LSun").value` at the end.

src/mcfacts/physics/lum.py
```diff
@@ -70,22 +84,71 @@ def shock_luminosity(smbh_mass, mass_final, bin_orb_a, disk_aspect_ratio, disk_d
     return Lshock
 
 
-def jet_luminosity(mass_final, bin_orb_a, disk_density, vk):
+def gas_capture_rate(
```
spacekaila: same comment here about using `.value` too early

src/mcfacts/physics/lum.py
```diff
@@ -70,22 +84,71 @@ def shock_luminosity(smbh_mass, mass_final, bin_orb_a, disk_aspect_ratio, disk_d
...
-    - disk_density: Gas density (kg m^3).
-    - vk: kick velocity (km/s).
+    ----------
+    mass_final : numpy.ndarray
+        mass of remnant post-merger (mass loss accounted for via Tichy & Maronetti 08)
+    bin_orb_a : numpy.ndarray
+        Orbital separation between the SMBH and the binary at the time of merger (in gravitational radii).
+    disk_density : callable
+        Function that returns the gas density at a given radius (in kg m^-3).
+    vk : numpy.ndarray
```
spacekaila: Please rename to `kick_velocity` or something like that so it's more clear!

src/mcfacts/physics/lum.py
```diff
@@ -95,40 +158,14 @@ def jet_luminosity(mass_final, bin_orb_a, disk_density, vk):
...
-
-    assert np.all(LBHL > 0), \
-        "LBHL has values <= 0"
 
-    return LBHL
+    #mcap = gas_capture_rate(bin_orb_a, disk_aspect_ratio, disk_density, mass_final, smbh_mass)
+    eta = spin_final**2
 
+    #LBHL = 1e42 * (mcap / 3e-4) * (eta / 0.5) * (0.1 / 0.1)
+    LBHL = 2.5e45 * (eta / 0.1) * (mass_final / 100)**2 * (vk / v_kick_scale)**-3 * (disk_density_cgs / 10e-10)  # Jet luminosity
```
spacekaila: Same comment here about units. If you define `mass_final_units = mass_final * u.Msun` and `kick_velocity_units = kick_velocity * u.km / (u.second ** -2)` you can use that as a final check that LBHL is in the right units.

Could also rename `LBHL` to be more clear, I assume it's `jet_luminosity`?

spacekaila (changes requested):
Added a bunch of comments! Some things (like the astropy units) are your choice whether to change or not.

src/mcfacts/objects/agnobject.py
emilymcpike: Added v_kick, lum_shock, lum_jet parameters to AGNMergedBBH attr list.

src/mcfacts/physics/analytical_velo.py
emilymcpike: Removed redundant np.array() declaration.

src/mcfacts/physics/analytical_velo.py
emilymcpike: Added comments regarding section and equation numbers to analytical velocity function

src/mcfacts/physics/analytical_velo.py
emilymcpike: Added astropy units to the analytical velocity function

src/mcfacts/physics/lum.py
emilymcpike: Followed through with units through the end and then grab .value

src/mcfacts/physics/lum.py
emilymcpike: Followed through with units through the end and then grab .value

src/mcfacts/physics/lum.py
emilymcpike: Renamed LBHL -> lum_jet and added units here

spacekaila (approved):

PR #298, opened 2025-05-01
Vectorized disk_capture.retro_bh_orb_disk_evolve and fixed bug

spacekaila:
#### Summary
- Vectorized the disk_capture.retro_bh_orb_disk_evolve function, which significantly sped up the code
- Fixed bug in calculations for the case where cos(w) = 0. We were dividing by the time, not the relevant scale factors. Fixed now.


xevra:
gw_strain
gw_strain
gw_strain

First image is from main-dev; second image is from this branch; third image is from this branch (before the bug fix)

xevra:
This confirms the bugfix is the reason for the increased number of mergers.

xevra (approved):

PR #307, opened 2025-05-09
Surrogate Model Implementation

shray4:
Surrogate model for mergers is now embedded into the mcfacts run process and generates more accurate remnant mass, spin, and kick velocity into the output_mergers_population.dat file


xevra:
Hi Shawn

Thank you for contributing a much better astrophysical model for kicks!
Please have patience, as I want to make sure that when these changes are integrated into the main-dev branch, that they don't break anything and other people can use them.
We're probably going to need to iterate a few times to accomplish this.

Major comments:
1. In general, it is bad practice to put code in the entry point of any Python code which is in the `/src` directory of a Python module. Everything in `/src` should be defined and never called, with certain exceptions (namely when a "main" function is defined and run `if __name__ == __main__: main()`

2. As a general rule, when you make substantial changes to a package, you should check that it can be installed without the "--editable" option. This should be followed with an attempt to install the package from a fresh conda environment (or one which has only had NumPy installed).

To test that all of the paths are working correctly, and that the code is as portable as it needs to be, it is also best to try testing the package using "python -m build" before distributing a package (or making a Pull Request)

These are especially important to do if you have:
- Created new Python package dependencies
- Created new data files that McFACTS needs in order to run
- Introduced modules of the code which are not purely in Python

In McFACTS, I created a "make build-install" and "make test-build" it's best to pip install without the "--editable" option. I want you to pass this test before I will approve the merge request.

3. Whenever you add something to the inifile, you need to:
- Update the hidden inifile
- Update ReadInputs to include the new arguments
- Make tests/inputs/test_ReadInputs.py work

On that last one, you'll need to update your branch by pulling from main-dev before it will work.

======== Minor Comments ========

At present, McFACTS doesn't run unless juliacall is installed.
pyproject.toml needs to be updated to reflect this so that pip will know to install juliacall.

At present, McFACTS doesn't run unless joblib is installed.
This should be reflected in the pyproject.toml

At present, McFACTS doesn't run unless scikit-learn is installed.
This should be reflected in pyproject.toml

At present, McFACTS doesn't run unless uncertainties is installed.
This should be reflected in the pyproject.toml

At present, McFACTS doesn't run unless quaternionic is installed.
This should be reflected in the pyproject.toml

At present, McFACTS doesn't run unless sxs is installed.
This should be reflected in the pyproject.toml

After installing these packages, make plots succeeds when installed with the --editable option.
The plots look fine.

At present, `make compare_sur` doesn't work. I do not believe compare_sure is included.

When running with `flag_use_surrogate = 1`, we get an error like
```
line 735, in load
    with open(filename, "rb") as f:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '../src/mcfacts/inputs/data/surrogate.joblib'
```

Instructions, and the need for files not currently present in McFACTS should be documented in the README, and in other places, including the pull request header.

======== Notes on the Diff ========

I will also go through and add my review on each file in the review.

Makefile: looks fine

recipes/model_choice_old.ini: looks fine. See summary discussion about ReadInputs.

removing_hash.py: This does not belong in the top level directory. Please put it in scripts.

scripts/binaryBHexp: Appears to be an empty file? Please explain what this is.

scripts/mcfacts_sim.py: What's going on with save_snapshots?

scripts/population_plots.py: Seems to work fine! :+1:

scripts/surfinbh_visualizations.py: Seems to be an empty file. Please remove or correct.

src/mcfacts/external/evolve_binary/evolve_binary.py: 
See my comment about the entry point in the summary. 
Currently, lines 9-10 run whenever the package is imported, which happens regardless of the inifile. 
That's the real problem here. 
One potential solution would be to only import this module when it is turned on in the inifile.
Generally, it is bad practice to import outside of the entry_point.
However, this is the exception: when only sometimes using a bunch of modules that most people have no business installing or setting up.
I will point this out in another section. Bug me if I don't.

src/mcfacts/external/evolve_binary/evolve_mcfacts.ipynb: Ipython notebook scripts do not belong in the src directory.
Please move to scripts, or create a "notebooks" folder at the top level directory for McFACTS and put it there.

src/mcfacts/external/evolve_binary/evolve_mcfacts.py: Too much is going on in the entry point here. 
Move to functions or create a script that has the desired behavior.

src/mcfacts/external/evolve_binary/example.py: See comment about entry points.

src/mcfacts/external/evolve_binary/fit_modler.py: Looks good.

src/mcfacts/external/evolve_binary/surrogate.py: I'm not sure what this is, but I'm fairly confident it doesn't belong in the src directory.

src/mcfacts/external/evolve_binary/test.txt: Does not belong in src

src/mcfacts/external/evolve_binary/vis.py: See comment about entry points.

src/mcfacts/external/manim/scene.py: Looks good

src/mcfacts/objects.agn_object.py: Why duplicate v_kick entry?

src/mcfacts/physics/binary/merge.py: merge_blackholes; is this code only executed when the inifile indicates that the surrogate model is to be used?

src/mcfacts/physics/evolve_bin.py: Looks good

recipes/model_choice_old.ini
xevra, line 1: See that this is added to the hidden inifile and the ReadInputs module as well

removing_hash.py
xevra, line 1: This does not belong in the top level directory. Please put it in scripts.

scripts/binaryBHexp
xevra, line 1: Appears to be an empty file? Please explain what this is.

scripts/mcfacts_sim.py
```diff
@@ -27,6 +27,9 @@
 from mcfacts.physics import point_masses
 from mcfacts.physics import lum
 from mcfacts.physics import analytical_velo
+from mcfacts.physics import evolve_bin
```
xevra: I would recommend moving these imports to `main()`, and only calling them if the inifile indicates surrogate model use.

scripts/mcfacts_sim.py
```diff
@@ -561,7 +564,7 @@ def main():
 
         while time_passed < time_final:
             # Record snapshots if user wishes
-            if opts.save_snapshots == 1:
+            if opts.save_snapshots == 0:
```
xevra: Is this a bug that you've fixed?

scripts/surfinbh_visualizations.py
xevra, line 1: Seems to be an empty file. Please remove or correct.

src/mcfacts/external/evolve_binary/evolve_binary.py
xevra, line 1: See my comment about the entry point in the summary.

src/mcfacts/external/evolve_binary/evolve_mcfacts.ipynb
xevra, line 1: Ipython notebook scripts do not belong in the src directory.
Please move to scripts, or create a "notebooks" folder at the top level directory for McFACTS and put it there.

src/mcfacts/external/evolve_binary/evolve_mcfacts.py
xevra, line 1: Too much is going on in the entry point here.  Move to functions or create a script that has the desired behavior.

src/mcfacts/external/evolve_binary/example.py
xevra, line 1: See comment about entry points.

src/mcfacts/external/evolve_binary/surrogate.py
xevra, line 1: I'm not sure what this is, but I'm fairly confident it doesn't belong in the src directory.

src/mcfacts/external/evolve_binary/test.txt
xevra, line 1: Does not belong in src

src/mcfacts/external/evolve_binary/vis.py
xevra, line 1: See comment about entry points.

src/mcfacts/objects/agnobject.py
xevra, line 1: Why duplicate v_kick entry?

src/mcfacts/physics/binary/merge.py
xevra, line 1: is this code only executed when the inifile indicates that the surrogate model is to be used?

xevra (changes requested):

xevra:
If you have any questions, I'm happy to help.

xevra (dismissed):
Very nice work Shawn!

PR #310, opened 2025-06-13
fixed bug that was not updating orb_a values for an if-statement

spacekaila:
#### Summary

The single-object dynamics functions (`circular_singles_encounters_prograde` and the extensions `_stars` and `_star_bh`) have an if-statement that checks if the orb_a of the circular object is within the min/max of the orb_a of the eccentric object, given its eccentricity. The orb_a of the circular object was stored in the `circ_prograde_population_locations = disk_bh_pro_orbs_a[circ_prograde_population_indices]` array, but when the orb_a of the circular object is actually updated, we update the `disk_bh_pro_orbs_a` array, not the `circ_prograde_population_locations` array. Thus, as the function continues, objects that have moved in the disk to no longer qualify for the if-statement will pass through.

**FIX**

I removed the `circ_prograde_population_locations` array so that we now only refer to orb_a values via using the `circ_idx` and `ecc_idx` indices on the `disk_bh_pro_orbs_a` array (or `disk_star_pro_orbs_a`) and everything is updated.


xevra (approved):

xevra:
Looks good!

PR #319, opened 2025-07-28
Proposing sweep function and adding tests

nrposner:
#Summary

Added proposed sweep variant for the circular_singles_encounters_prograde function, along with testing infrastructure for it.

## Details

The implementation of the sweep function added here contains both the sweep algorithm and the original double-loop. It performs a runtime check to determine whether or not the overall length and proportion of C and E makes it likely that the sweep function will deliver improved runtime performance. If yes, it proceeds with the sweep function, otherwise falls back to the double loop.

There is a single test covering a range of N and C/E proportions. The tests ensure that, for each run, the outputs are of the same length and all elements are identical to within stated tolerances (eccentricities are only accurate to within 10e-4, while axes are accurate to within 10e-7. We also check the runtime performance of both the original and sweep functions to ensure that none of them saw a >20% slowdown. Since the runtime performance of the function is less predictable as N gets smaller, and it is precisely at small N (<100) where the double loop still outperforms the sweep function, we cannot get substantially more precise without just causing the test to fail randomly because of background noise.

The test ran green 20 times in a row on the local environment before this push.


scripts/tests/encounters_optimization_tests.py
```diff
@@ -0,0 +1,369 @@
+import time
+import numpy as np
+import pytest
+
+import astropy.units as u
+import astropy.constants as const
+
+from mcfacts.mcfacts_random_state import rng
```
xevra: You import this, but don't use it. As this script is now, I would comment it to avoid confusion.

scripts/tests/encounters_optimization_tests.py
```diff
@@ -0,0 +1,369 @@
...
+# from mcfacts.physics.dynamics import circular_singles_encounters_prograde, circular_singles_encounters_prograde_sweep
```
xevra: I'm going to be a bit of a stickler about this.

Right now you are testing functions called "circular_singles_encounters_prograde" and "circular_singles_encounters_prograde_sweep", but you are not testing the functions in the McFACTS library.

If there was a typo, or if one were changed and not the other in a future update, it would become possible for the test to "pass" while the function we rely on becomes broken.

You should be testing the functions in the library. They should be defined there and not here.

Additionally, you define all of these tests, but pytest does not currently run them, because there is no `if __name__ == "__main__": do this` block at the bottom of this script.

Additionally, this script should be moved to the `tests` folder in the top level directory of McFACTS.

scripts/tests/encounters_optimization_tests.py
```diff
@@ -0,0 +1,369 @@
...
+def generate_data(N: int, circ_proportion: float, rng: np.random.Generator):
+    """Generates random mock data for the simulation functions."""
+    # Mock physical constants
+    mock_params = {
+        "smbh_mass": 1.0e8,
+        "timestep_duration_yr": 1.0e4,
+        "disk_bh_pro_orb_ecc_crit": 0.1,
+        "delta_energy_strong": 0.11, # setting it to greater than disk_bh_pro... to avoid bug
+        "disk_radius_outer": 20000.0
+    }
```
xevra: This is fine. It might be better to read them from the default inifile at some point in the future, but none of our other tests do that yet (except the test for ReadInputs), so don't sweat it.

scripts/tests/encounters_optimization_tests.py
```diff
@@ -0,0 +1,369 @@
...
+    # --- Run Optimized Function ---
+    data_for_opt = {k: v.copy() if isinstance(v, np.ndarray) else v for k, v in data.items()}
+    start_time = time.perf_counter()
+    a_opt, ecc_opt = circular_singles_encounters_prograde_sweep(**data_for_opt, rng=rng2)
+    time_opt = time.perf_counter() - start_time
+    print(f"Optimized took:  {time_opt:.4f} seconds")
+
+
+
+    # --- Correctness ---
```
xevra: Ideally, these tolerance values would be identified at the top of the script and named appropriately.

xevra (changes requested):
Looks good so far! Looking forward to seeing the CI tests test the library functions.

scripts/tests/encounters_optimization_tests.py
nrposner: I'm a bit confused here. My impression was that using pytest, we just needed to make sure all our function names start with test_, and put them in files whose names start with test_ or end with _test. Then we just run python3 -m pytest, and it discovers and runs all test functions in all test files. Are we running the tests from scripts rather than from pytest directly?

scripts/tests/encounters_optimization_tests.py
nrposner: switched to using that rng import and reseeding it as necessary

scripts/tests/encounters_optimization_tests.py
nrposner: redefining tolerances at the top of the script

nrposner:
Okay, pretty sure the failed tests here aren't mine...

xevra:
> Okay, pretty sure the failed tests here aren't mine...

I'll restart the tests. Something strange may have happened.

However, I know the test we ran yesterday on the same pull request passed the test, so there must be something going on.

I'll try running the tests on my laptop.

xevra:
I figured out what happened.

It's due `setupdiskblackholes` calling the random number generator and getting different random numbers because the encounters test happened first.

I suspect the problem has something to do with scope.

You can avoid this by using your own np.random.RandomState() object instead of importing the one from mcfacts.

I tested it on my laptop that way, and the tests passed.

xevra:
This time your test failed because it's not faster, but that's actually okay. We don't want to stress test the N-body portion of the code in the unit test. Please comment out the slowdown assertion.

xevra:
I think the CI is a poor environment for a performance test. We'll notice the speed improvement when we are running mcfacts for sure.

xevra (approved):

PR #320, opened 2025-07-30
dynamics: added flag_dynamics_sweep

xevra:
#### Summary

dynamics: added flag_dynamics_sweep option to use new circular_singles_encounters_prograde_sweep function

#### Issue Link

Changes the default behavior of model_choice_old.ini to use the new sweep implementation created by Nico, and provides a way to turn it on/off (flag_dynamics_sweep).

Edit: See https://github.com/McFACTS/McFACTS/pull/319

### DIFF
--- a/IOdocumentation.txt
+++ b/IOdocumentation.txt
@@ -233,6 +233,9 @@ star cluster variables second, and assorted other choices (time resolution etc.)
         default: 0.2735
     nsc_star_metallicity_z_init : float
         default: 0.02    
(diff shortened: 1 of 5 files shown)


xevra:
By the way, it is one second faster on my M1 (1 minute, 17 seconds instead of 1 minute 18 seconds) for the fiducial model. However, as I prepare to start the larger runs, I expect the improvement to be more significant.

spacekaila (approved):

PR #327, opened 2025-08-06
fix [phys]: fixing circular_singles_encounters_prograde edge check

nrposner:
Fix the edge case error which occurs when a particle is exactly on the edge of the disk radius, such that it is not kicked back in yet does not pass the assert test.

### DIFF
--- a/src/mcfacts/physics/dynamics.py
+++ b/src/mcfacts/physics/dynamics.py
@@ -671,7 +671,7 @@ def circular_singles_encounters_prograde(
                             disk_bh_pro_orbs_ecc[circ_idx] = delta_energy_strong
                             disk_bh_pro_orbs_a[circ_idx] = disk_bh_pro_orbs_a[circ_idx]*(1.0 + delta_energy_strong)
                             # Catch for if orb_a > disk_radius_outer
-                            if (disk_bh_pro_orbs_a[circ_idx] > disk_radius_outer):
+                            if (disk_bh_pro_orbs_a[circ_idx] >= disk_radius_outer):
                                 disk_bh_pro_orbs_a[circ_idx] = disk_radius_outer - epsilon[i][j]
                             disk_bh_pro_orbs_ecc[ecc_idx] = disk_bh_pro_orbs_ecc[ecc_idx]*(1 - delta_energy_strong)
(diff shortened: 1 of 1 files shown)


src/mcfacts/physics/dynamics.py
```diff
@@ -791,7 +791,7 @@ def circular_singles_encounters_prograde_sweep(
                         # apply state change, using the fixed logic
                         disk_bh_pro_orbs_ecc[circ_idx] = delta_energy_strong * 1.0001
                         disk_bh_pro_orbs_a[circ_idx] *= (1.0 + delta_energy_strong)
-                        if (disk_bh_pro_orbs_a[circ_idx] > disk_radius_outer):
+                        if (disk_bh_pro_orbs_a[circ_idx] >= disk_radius_outer):
```
xevra, line 794: Should it not be:
`if (disk_bh_pro_orbs_a[circ_idx] >= disk_radius_outer -  epsilon[circ_rel_idx][ecc_rel_idx])`?

nrposner: I figured we want to test whether the particle is exactly on the edge of the disk radius (the `disk_bh_pro_orbs_a[circ_idx] == disk_radius_outer` case) and include that in the conditions to kick the particle back in, no?

nrposner:
Further debugging has shown that the issue is as follows: in the input to circular_singles_encounters_prograde at seed 3456791654 there is one element at index 107 which is exactly equal to the disk radius, and which does not experience any encounters which would adjust it downwards, hence why changing to >= checks doesn't solve the issue.

The simplest 'fix' for this would be to change the assertions to assert that any value is less than or equal to the disk radius, accepting that there could be values which are exactly on the edge of the disk radius. 

On the other hand, we could choose to adjust any values found to be exactly on the edge down by a tiny amount up front. However, this could lead to different values in unexpected places.

xevra (approved):
So, this is a decision we need to make as a team. We either want to:
- allow orbits to have a semi-major axis equal to the disk radius and change the assert statement
OR
- not allow this, and figure out which part of the code is setting it identically equal to the disk radius and fix it there.

In other words, this isn't a bug that you created.

Therefore, I'm going to approve the PR because this is arguably an improvement and doesn't break anything.

PR #329, opened 2025-08-06
Created new wrapper class for rng

spacekaila:
#### Summary

Created a new wrapper class for our random number generator (rng). It inherits all the attributes and classes of the previous rng approach we were using, but now there is an additional attribute `.call_count` that keeps track of how many times the rng is called.

WARNING: it is a bit slower on my machine. The previous approach took 1:23 for a default run, this new approach took 1:37.

I also added a few newlines in the output, as it's hard to read when all the galaxy results are crammed together.


xevra (changes requested):
This is all very Pythonic, and is good work. However, it doesn't quite solve the problem we discussed last week.

Rather than incrementing by 1, you need to increment by the `size` keyword argument when given, and incrementing by 1, otherwise.

Example demonstrating the issue:

```
In [6]: import numpy as np

In [7]: rs1 = np.random.RandomState(42)

In [8]: rs2 = np.random.RandomState(42)

In [9]: rs1.normal()
Out[9]: 0.4967141530112327

In [10]: rs2.normal()
Out[10]: 0.4967141530112327

In [11]: rs1.normal(size=10)
Out[11]: 
array([-0.1382643 ,  0.64768854,  1.52302986, -0.23415337, -0.23413696,
        1.57921282,  0.76743473, -0.46947439,  0.54256004, -0.46341769])

In [12]: rs2.normal(size=5)
Out[12]: array([-0.1382643 ,  0.64768854,  1.52302986, -0.23415337, -0.23413696])

In [13]: rs2.normal(size=5)
Out[13]: array([ 1.57921282,  0.76743473, -0.46947439,  0.54256004, -0.46341769])
```

JakePost:
@xevra does [b179620](https://github.com/McFACTS/McFACTS/pull/329/commits/b179620aabf74d6db780bf3c536f48adc812a269) address what you requested to be changed?

JakePost:
I 'upgraded' the random generator to the [Philox algorithm](https://numpy.org/doc/stable/reference/random/bit_generators/philox.html).

Philox is a counter-based PRNG, which allows us to just probe the state of the bit generator to find the number of calls.

_Note: This number does not directly correlate to the number of functions called on the generator. This is especially true when pulling from a distribution as the underlying function progresses the bit generator state multiple times when instantiated._

JakePost:
Was it decided that this wouldn't go into main-dev and to wait for this feature in restructure?

If so, should this be closed?

xevra:
@JakePost I think you had volunteered to work on this back in October, but that was a long time ago (before the PCG64 / other random number generator discussion). Therefore, I don't think this pull request has gotten a lot of attention. Maybe we can discuss it? I don't think it's going to be in 0.4.0, but there could always be a 0.4.1 (and smaller PRs to main is a good idea actually).

PR #336, opened 2025-08-12
EM models for LSST

varunprit:
#### Summary

Added a script to calculate the EM model of a the thermal emission of a BBH merger in AGN disk with the velocity of forward shock. Also added data files for LSST filters that are being used here


scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+import pandas as pd
+from astropy.io import fits
+from astropy.constants import h, c, k_B, m_p
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
+import mcfacts
+
+# Constants used 
+#t = 5800 #Temp of sun for now can change for what we are looking at 
+rad = 6.96e8 #Radius of sun in meters for now will change to bh later
```
cookhe, line 11: Please use `astropy.constants.R_sun`

varunprit: Its there as more of a place holder to be whatever the radius of the BH merger would be so i didnt use the constant. Would it be better to just comment it out?

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+d = 1.5e11 #distance from earth to sun in meters
```
cookhe: Please use `astropy.constants.R_sun` and `astropy.constants.au` in case it's

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+#t = 5800 #Temp of sun for now can change for what we are looking at 
+rad = 6.96e8 #Radius of sun in meters for now will change to bh later
+d = 1.5e11 #distance from earth to sun in meters 
+h = h.value #Planck constant
+c = c.value #speed of light
+k = k_B.value #Boltzmann constant
+m_p = m_p.cgs.value #proton mass
+pi = np.pi #Pi constant
+v_fs = 0.1 * c #velocity of forward shock (used in Beta)
+dens = 1e-15 #density of agn disk
```
cookhe: Since `v_fs` and `dens` are parameters that the user can/should interact with, it is better to place them immediately following the imports and separated from other code. That way a user doesn't need to hunt for the parameters they need to change. You can also leave the definition for `v_fs` here since it relies on `c` and substitute the `0.1` value with an object named `c_factor`. Then, set `c_factor=0.1` above next to `dens`.

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+i_freq = c/i_wave
+u_freq = c/u_wave
+y_freq = c/y_wave
+z_freq = c/z_wave
+
+
+b_fs = v_fs / c
+gamma_fs = 1 / np.sqrt(1 - b_fs**2)
+
+def temp_fxn(v_fs, dens, m_p, c, k):
```
cookhe, line 25: It's standard Pythonic procedure to define functions together at the beginning of your script. 

Please place this function, and the two you define below, after the section you create for user-controlled objects and before you define all of the constants.

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    Calculate the temperature t based on velocity of forward shock and other physical parameters.
+
+    Parameters
```
cookhe, line 29: It might be helpful to someone using the function out of context to know what units these parameters should have (e.g., m/s, cm/s, K, etc.). Likewise for the other functions' parameters and outputs. Please add where relevant. A "all parameters should be in MKS/CGS units" statement in the function description would cover this nicely if that is indeed the case.

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+t = temp_fxn(v_fs, dens, m_p, c, k)  # Calculate temperature based on forward shock velocity and density
+
+def planck_fxn(filter, t):
+    """Planck's law for frequency
+    This function calculates the spectral radiance of a black body at a given temperature T for a specific frequency filter.
+    
+    
+    Parameters
+    ----------
+    filter (array-like): The frequency values for which to calculate the spectral radiance.
```
cookhe: 1. move this function up to the beginning of the script according to my comment above.
2. Please format these parameter entries according to the numpy docs standard:
```
parameter : type
    Description
```

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+total_freq = np.concatenate((r_freq, g_freq, i_freq, u_freq, y_freq, z_freq))#sorting the frequency values
+sorted_freq = np.sort(total_freq)
+
+# To make plancks law graph smooth 
+total_index = np.argsort(total_freq)
+total_wave = total_wave[total_index]
+planck_total = (planck_total[total_index])  
+
+# planck law based on area
+planck_area = (planck_total * pi * (rad*2))
```
cookhe: Should this be `rad**2`? I'm interpreting this equation as `P_a = P_t * (2 pi R)`. But `2 pi R` is circumference not area. However, this may not be important because `planck_area` is only used to calculate `planck_disk` which is not used anywhere else. Are these parameters planned to be used in the future by some more additions to the script?

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+
+# planck law based on distance 
+planck_dist = (planck_area/(4 * pi * (d**2)))
+
+#t_mid = 1e4
+#for when b is greater than 0.03 but less than 1 use wien
+def wien_fxn(filter, T = t):
```
cookhe: Move to the beginning of the script according to my comment above.

scripts/EM_plot.py
```diff
@@ -0,0 +1,256 @@
...
+wien_i = wien_fxn(i_freq) 
+wien_u = wien_fxn(u_freq)
+wien_y = wien_fxn(y_freq)
+wien_z = wien_fxn(z_freq)
+
+wien_freq = np.concatenate((wien_r, wien_g, wien_i, wien_u, wien_y, wien_z)) #combining all the wien values
+wien_total = (wien_freq[total_index])  # sorting the wien values 
+
+# wien law based on area
+wien_area = (wien_total * pi * (rad*2))
```
cookhe: Same comment as for `planck_area` above.

cookhe (approved):
Approved. Nice work!

PR #338, opened 2025-08-16
Saving immortal star files, update parameter for star coalescence

spacekaila:
#### Summary

- Created arrays `stars_all_id_nums`, `stars_masses_initial`, and `stars_orb_a_initial`. Every time a new star is made (at galaxy initialization, mergers, or NSC capture), the new star's parameters are added to these arrays.
- Created new `AGNImmortalStar` object that has all the parameters of `AGNStar`, but also adds `mass_initial` and `orb_a_initial`.
- After stars accrete mass, McFACTS checks for immortal stars. If any exist, we check their IDs against the `stars_all_id_nums` array and grab their original `mass` and `orb_a` values. Add current parameters and initial parameters to `AGNImmortalStar` object.
- `AGNImmortalStar` is saved to file at end of galaxy like other star files.
- Changed value of `Pr` for the coalescence option for stars (turned off by default). It was set to 1, should (and is now) set to 2 as it is the powerlaw index for the star `orb_a` distribution.
- Added print statement for if `flag_coalesce_initial_stars` is turned on that prints the max stellar mass post-coalescence and the number of stars over `disk_star_initial_mass_cutoff` (immortal limit).
- Changed line for new ID numbers for BBH that break apart at the end of the galaxy to ensure that the ID numbers in the `blackholes_pro` and `filing_cabinet` objects are the same.


spacekaila:
I also commented out the lines calculating GW frequencies for "TDEs" because we don't treat TDEs in any physically consistent way and it is constantly bugging out and breaking McFACTS. So until I/someone has the time to make it work properly I just turned it off.

cookhe (approved):
The immortal stars are unique in each output file. Tested for 5 Galaxies.
Verified BH figures (no stars) for 100 galaxies look similar to the benchmarks.

PR #342, opened 2025-10-07
[perf]: Eliminating unnecessary operations from si_from_r_g

nrposner:
At present, the si_from_r_g function in point_masses.py performs a large number of unnecessary operations by recalculating the r_g value from scratch each time it is called. Since the function is called around 100,000 times in the course of a default sim.py run, this cost balloons to account for nearly a fifth of total runtime (profiled on the current version of main_dev). 

We identify that, first, each function call is unnecessarily heavy due to the use of unit conversion code to create the SI units, and second, that the base r_g calculated each time is not expected to change during a run, as it is created from astronomical constants and the SMBH mass.

Thus, we advocate rewriting si_from_r_g to optionally accept a pre-computed value of r_g in meter units, allowing the function to skip the expensive initialization of this value in most cases. In order to enable this, we suggest calculating an r_g_in_meters value at initialization time and adding it to the values dictionary created by `ReadInputs_ini`, so that it can be passed as an argument into top-level functions as `opts.r_g_in_meters`. 

This has been implemented in main_dev_si on the remote fork, and shows a significant performance gain: on a local CProfile run of sim.py, total runtime drops from 222 seconds to 184 seconds, with 31 seconds of that 38 second decrease being directly attributable to the improvement in performance of the si_from_r_g function, which goes from accounting for 43 seconds of runtime to 12 seconds. 

This amounts to an elimination of 13% of runtime, or a 20% runtime performance improvement. 

Note that this run did not use the new kwarg for si_from_r_g calls in the `merge` or `lum` modules. Expanding the optimization to those modules as well would likely remove most of the remaining 12 seconds of runtime for this function, likely amounting to a total runtime performance improvement in excess of 25%. 

Before:
Screenshot 2025-10-07 at 7 21 07 PM

After:
Screenshot 2025-10-07 at 7 21 26 PM

---

## Notes

---

This PR is split across a large number of files, since it is necessary to not only adapt the si_from_r_g function (point_masses.py) and to add the precomputed quantity to opts (ReadInputs.py), but also to add it to the function parameters for a large number of functions which call si_from_r_g (many files in physics/) and update the top-level function calls to include it (sim.py). 

---

Note that there remain some unresolved issues that must be addressed prior to merging:

When running make plots, the script ends seemingly successfully, but with the following warning:

```
Performance time: 53.76  [ 0 min - 53 sec ]
cd runs; \
        python .././/scripts/population_plots.py --fname-mergers .//output_mergers_population.dat --plots-directory ./
.//output_mergers_population.dat
~~~~~~~~~~~~~~~~~~~~~~
 [ReadOutputs.py] Warning!: The log file you're using contains 1 additional entries not found in OUTPUT_TYPES. They have been added to the log dictionary as a STRING type. Please verify their intended types before you use them in your analysis or add them to the INPUT_TYPES or OUTPUT_TYPES dictionaries.
  r_g_in_meters: 147662503805.01248 m
~~~~~~~~~~~~~~~~~~~~~~
/Users/nicol/rusty/projects/McFACTS/runs/.././/scripts/population_plots.py:679: UserWarning: The figure layout has changed to tight
  plt.tight_layout()
Warning: Could not find transfer function file!
                Approximation will be used...
```

This likely arises from the addition of the r_g_in_meters value into the `ReadInputs_ini` values dictionary. I do not know what the consequences of this warning are or how to exclude r_g_in_meters from being included in the logs, and request guidance on this matter. 

In addition, while `make tests` runs without issue, I'm not sure if there are other scripts or tests that I should be running to ensure the changes are not breaking other assumptions. I also request guidance on this matter.


mstar.prof
cookhe, line 1: I could not find a reference to this file in the code. What is it, and why is it included in the PR?

Edit: After some research, I think these `*.prof` files are output files from the cProfile python module. Do we need them for McFACTS? If not, I believe it should be removed from the PR, but maybe I'm not aware of its purpose.

optimized.txt
cookhe, line 1: What is this file? It is a binary file not a `*.txt` file.

nrposner: This and the others are leftover profiles, removing now.

src/mcfacts/inputs/ReadInputs.py
```diff
@@ -228,6 +229,27 @@
     raise ValueError("[ReadInputs.py] Boolean data types are not allowed in"
                      "the INPUT_TYPES dictionary. Please use int instead.")
 
+
+def initialize_r_g(input_variables):
+    # pre-calculating r_g from the provided smbh_mass
+    # Initialize the shared value to None
```
cookhe, line 249: Please add a docstring that follows the format used in McFACTS.

src/mcfacts/inputs/ReadInputs.py
```diff
@@ -228,6 +229,27 @@
...
+    G = ct.G.to('m^3/(kg s^2)')
+    # Assign units to smbh mass
+    if hasattr(input_variables["smbh_mass"], 'unit'):
+        smbh_mass = input_variables["smbh_mass"].to('solMass')
+    else:
+        smbh_mass = input_variables["smbh_mass"] * u.solMass
+    # convert smbh mass to kg
+    smbh_mass = smbh_mass.to('kg')
+    # Calculate r_g in SI
+    r_g_in_meters = G*smbh_mass/(c ** 2)
```
cookhe: Linting/style suggestion:
```suggestion
    r_g_in_meters = G * smbh_mass / (c ** 2)
```

src/mcfacts/inputs/ReadInputs.py
nrposner: Docstring added

mstar.prof
nrposner: These were detritus, shouldn't have left my local environment. All removed now.

src/mcfacts/physics/point_masses.py
```diff
@@ -151,7 +151,7 @@ def orbital_separation_evolve_reverse(mass_1, mass_2, sep_final, evolve_time):
     return sep_initial * u.m
 
 
-def si_from_r_g(smbh_mass, distance_rg):
+def si_from_r_g(smbh_mass, distance_rg, r_g_defined = None):
```
cookhe: Another formatting suggestion.
```suggestion
def si_from_r_g(smbh_mass, distance_rg, r_g_defined=None):
```

.gitignore
```diff
@@ -84,3 +84,4 @@ coverage.xml
 cover/
 
 docs/source/_autosummary
+si.prof
```
cookhe: What process creates this file? Maybe left over from the other `*.prof` files?

si_optimized.prof
cookhe, line 1: Is this used for anything?

.gitignore
nrposner: Accidentally added it to gitignore, removed now

si_optimized.prof
nrposner: Nope, removed

cookhe:
This error 

```
~~~~~~~~~~~~~~~~~~~~~~
 [ReadOutputs.py] Warning!: The log file you're using contains 1 additional entries not found in OUTPUT_TYPES. They have been added to the log dictionary as a STRING type. Please verify their intended types before you use them in your analysis or add them to the INPUT_TYPES or OUTPUT_TYPES dictionaries.
  r_g_in_meters: 147662503805.01248 m
~~~~~~~~~~~~~~~~~~~~~~
```
can be solved by adding `r_g_in_meters` to the `INPUT_TYPES` dictionary found in `ReadInputs.py`. This dictionary is used to ensure proper type casting throughout the code. 

The warning message arises because the log file is generated from `opts`, so any new values added to opts at setup will exist there. Then, when plotting or performing other analysis, `ReadOutputs.py` imports both `INPUT_TYPES` and `OUTPUT_TYPES`  to properly typecast the log file values. If the log file now contains a line that does not match any item in those dictionaries, the script assigns a string type and warns the user to that fact.

No issues will arise except if the analysis assumes the new line item is some other type and the developer didn't verify the values after loading the log file - hence the warning and not an exception. In this case users might reasonably anticipate `r_g_in_meters` is a float, so it would be beneficial to add it to `INPUT_TYPES` (since it's being used throughout the code and isn't something calculated or added at the end of a McFACTS run).

cookhe (approved):
Plots match the current `main-dev` set.

Performance times:
  1. `main-dev` - 118.36 s
  2. this PR - 101.19 s
Speedup ~= 15%

PR #350, opened 2025-11-10
Added option to calculate orbital eccentricity and inclination

varunprit:
Changed the merged_orb_ecc function to merged_orb_ecc_and_inc so it includes the new bh_kick_comp_merged from shawn. Added in the function the ability to calculate both the eccentricity and inclination in the same function as well so it outputs both.


JakePost:
On hold for the following:

1. Decided that this won't go into stable version 0.4.0 #357 
2. Discussion about what else needs to go into this PR (if anything)

JakePost:
Should be re-implemented for latest version of main-dev.

PR #351, opened 2025-12-01
Optimization: Accelerating tau_*_dyn functions with Rust + Pedantry

nrposner:
Noting that half the time (when `retro_bh_orb_disk_evolve` is called from line 1043 of mcfacts_sim.py), the type of retrograde_bh_masses is `float` rather than `ndarray[float]`. This can be verified by adding

```python
print("Type of retrograde_bh_masses is:", type(retrograde_bh_masses))
print("Type of retro_mass is:", type(retro_mass))
```

into `tau_semi_lat()` after the definition of retro_mass. 

The input in these cases is `stars_retro.mass`. If this is the intended behavior, then we should correct the type notes in use here. 

If this is not the intended behavior and `stars_retro.mass` is always intended to be an `ndarray[float]`, not a `float`, then this should be addressed


nrposner:
ANOTHER Julia error???

nrposner:
This is coming from the same head repo, so adding it here:

We introduce the first set of Rust accelerants for McFACTS, as part of the [McFAST](https://pypi.org/project/mcfast/) helper module, now on PyPI. 

The goal of this function, and similar accelerant functions to come, is to provide alternative 'optimized' functions that have identical call signatures and identical outputs to some of our heavier functions, but work much more quickly. In the current build of main-dev, `retro_bh_orb_disk_evolve` accounts for 13.5% of total runtime, our second heaviest function call, and that in turn is dominated by calls to `tau_ecc_dyn` and `tau_inc_dyn`, accounting for 7.9% and 4.6% of total runtime respectively.

McFAST surfaces two functions, `tau_ecc_dyn_helper`, `tau_inc_dyn_helper`, which take over the vectorized computations performed by NumPy with optimized Rust functions. As part of this PR, we also produce two new Python-side functions, `tau_ecc_dyn_optimized` and `tau_ecc_dyn_optimized`, which can be dropped in where the original functions are currently used, and will produce the same output: they perform some up-front data transformation in Python where relevant (mainly calling si_from_r_g), and then deploy the Rust helper functions.

In testing, this results in a consistent global runtime reduction: total runtime reduces from 126 seconds to 104 seconds, an 18% reduction. 

Screenshot 2026-01-17 at 4 19 04 PM

Screenshot 2026-01-17 at 4 19 19 PM

Half that reduction can be attributed directly to this intervention, with `retro_bh_ob_disk_evolve` going from 18.0s to 7.2s in runtime, a 60% reduction, as a result of calling the optimized function variants rather than the originals. The vectorized calculations themselves see even stronger speedups, with the `tau_semi_lat` function that we originally timed at 6.4 seconds being subsumed entirely into `tau_ecc_dyn_helper`, which is now timed at 0.06 seconds, a >100x speedup.

The other half of this reduction seems to arise from scattered time reductions across other major functions, for which we cannot properly account: this may be the result of variances and instability in cProfile, but the further reductions appear to be consistent across multiple runs. We encourage reviewers to double check that the outputs of the simulation do not change as a result of the optimized functions. 

We also performed local checks for stability, currently commented out: np.allclose() confirms that paired outputs were identical, having the same dimensions and all contents being equal to within 5 decimal points precision.

We also include pedantic documentation changes to highlight that, in current usage, both tau_ecc_dyn and tau_inc_dyn are supplied with floats and arrays of floats in equal proportion for the argument `retro_bh_orb_mass` and equivalent.

In order to pass muster, the [upstream code](https://github.com/nrposner/mcfast/blob/main/src/accelerants/tau.rs) in McFAST itself should also be reviewed by a second reviewer familiar with Rust (ie Vera). 

Should this PR be approved, we suggest that McFAST be added as an official dependency of the McFACTS project, and that the main McFAST repo and PyPI page move to the ownership of the McFACTS github account for ease of future maintenance.

All benchmarking was performed on an M3 Max on High Power with cProfile running in parallel.

nrposner:
Added testing file for this change at `tests/tau_dyn_test.py`. Uses actual sampled input data, and reconstructs one of the possible density lambdas to use as a default. Currently runs and confirms that all outputs between the original and optimized functions are equal to within a relative tolerance of 1e-9, which I think should be enough.

Also added mcfast >= 0.1.4 to the pyproject.toml dependency.

nrposner:
Restored mcfacts support for intel macs.

xevra (approved):
`make plots` runs with time:
```
Performance time: 41.70  [ 0 min - 41 sec ]
```
on my machine (M4 Pro).

No issues installing mcfast from pypi.

Test passes locally, and tests new functions effectively.

PR #352, opened 2025-12-12
Retrograde Math Fix

JakePost:
- Correcting the order of operations issue in the tau_semi_lat method slows down retrograde evolution.

- Correcting the check for things flipping prograde to only look at the orbital inclination. It was previously setup to let anything with zero eccentricity flip.

- Setting the lower limit of retrograde eccentricities to the critical eccentricity (0.01). Having an eccentricity of exactly 0 causes some issues with the analytical model. Since retrogrades were incorrectly flipped, this was previously not an error we ran into.

To mitigate the red storm, we can artificially reproduce the interaction we think the retrograde dump had with the eccentric population. To do this, Barry and I raised the max eccentricity from 0.3 to 0.9 and raised the number of timesteps from 50 to 70.

### DIFF
--- a/recipes/model_choice.ini
+++ b/recipes/model_choice.ini
@@ -80,7 +80,7 @@ disk_bh_torque_condition = 0.1
 #   Eddington accretion rate
 disk_bh_eddington_ratio = 1.0
 # Maximum initial eccentricity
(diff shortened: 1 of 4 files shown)


cookhe (approved):
Verify that the corrections produce plots similar to the diagnostic set.

PR #354, opened 2026-01-26
Renamed exploded stars to disrupted

spacekaila:
#### Summary
Renamed all variables with some version of `star_exploded` to `star_disrupted`. Added `disk_sound_speed` to arguments in `merge_blackholes` function call on line 1685 in `mcfacts_sim` because it had been omitted.


xevra (approved):
Looks good, and runs locally with no issues!

Issue #347, opened 2025-10-15
Pull Request CI is broken


xevra:
The last successful CI run was: https://github.com/McFACTS/McFACTS/actions/runs/18062099405

The issue does not appear to be the python environment, as I ran a test [PR 346](https://github.com/McFACTS/McFACTS/pull/346) with the exact same Python environment ([See pyproject.toml for PR 346](https://github.com/McFACTS/McFACTS/blob/83a84d30455aa053a8673be6f91a89b4b69cb491/pyproject.toml#L66)) as the [previous successful run](https://github.com/McFACTS/McFACTS/actions/runs/18062099405/job/51399812928#step:4:476), and it still failed.

I don't know what is causing this or how to fix it.

One potential solution is to take more of the SxS/Julia stuff and handle the dependencies differently, requiring users to install them manually if they want to use the NR Surrogate. That's not a good solution, but it might work in the short term.

None of the current PRs will pass tests unless this is fixed, even those having nothing to do with the SxS approximant surrogates.

PR #355, opened 2026-01-28
Nojulia

xevra:
#### Summary
Removed sxs (and julia) as a direct dependency. The code will now install and run without the sxs module or any julia. However, it will still try to import those packages if and only if the user has set the `flag_use_surrogate = 1` in the inifile during the inputs parsing phase.

Also added python and numpy versions to prevent accidentally running into issues because of version that the developers have not tested on.

Edit: nrposner contributed and fixed pytest integration for new unit tests with mcfast.

Edit2: Addresses this Issue: https://github.com/McFACTS/McFACTS/issues/347


xevra:
This test is failing because pytest is trying to interface with the `tau_dyn_test.py` test that Nico set up.

The test passed for me on that PR when I ran it from the terminal, but doesn't run in pytest.

I didn't try pytest or the CI on that branch because julia was breaking everything.

xevra:
pytest still runs locally and in CI after merging from main-dev into nojulia.

JakePost:
nojulia -> main-dev: all local tests pass and diagnostic graphs show the expected outputs for a default run and surrogate run.

JakePost (approved):

PR #361, opened 2026-02-02
Paper 4 housekeeping

emilymcpike:
### Summary of Changes
 Changed paper_em name for .ini file folder to paper_4, reverted phenom_turb_std_dev = 1.0 to 0.1 (default=0.1).

### Validation Steps
make plots
make em-plots

### Checklist

- [X] Commit messages follow the Conventional Commits guidelines
- [X] Added / updated documentation


recipes/model_choice.ini
```diff
@@ -37,7 +37,7 @@ flag_phenom_turb = 0
 #Centroid of Gaussian w.r.t. to migrating BH. 0 (default)
 phenom_turb_centroid = 0.
 #Variance of Gaussian around Centroid (default=0.1)
-phenom_turb_std_dev = 1.0
+phenom_turb_std_dev = 0.1
```
xevra, line 40: This changes the default inifile for McFACTS. Please explain this change in the context of how it might affect our outputs.

src/mcfacts/setup/setupdiskblackholes.py
```diff
@@ -191,6 +191,11 @@ def setup_disk_blackholes_masses(
         # paper: 2023PhRvX..13a1048A, Section VI.B
         while (np.sum(disk_bh_initial_masses > nsc_bh_imf_max_mass) > 0):
             disk_bh_initial_masses[disk_bh_initial_masses > nsc_bh_imf_max_mass] = rng.normal(loc=mass_pile_up, scale=2.3, size=np.sum(disk_bh_initial_masses > nsc_bh_imf_max_mass))
+    elif nsc_imf_bh_method == "peak":
```
xevra: Please describe what the "peak" imf method is supposed to draw from in words, in the PR header.

recipes/model_choice.ini
emilymcpike: I had previously used this to test the phenomenological turbulence mod, which doesn't effect anything in default. This reverts the value to the default value as documented in the .ini file

src/mcfacts/setup/setupdiskblackholes.py
emilymcpike: Updated, is this sufficient?

xevra: I think if your expected behaviour is that it will draw all of the black hole masses from the 35 solar mass peak bump, there is a bug in this method.

The pareto distribution should not be run at all in this case. (your line 195).

Also, there is already a "gaussian" IMF reference, which can have a peak and standard deviation defined in the inifile. 

This is a duplicate feature, and it has a major bug.

xevra (changes requested):
This is a duplicate feature and contains a major bug (see comments).

xevra (dismissed):

xevra (approved):

PR #367, opened 2026-02-17
New optimized function variants with mcfast

nrposner:
We add a new mcfast version, version 0.1.6, with five new helper functions for seven McFACTS functions.

- gw_strain_helper (gw_strain_freq_optimized)
- merged_orb_ecc_helper (merged_orb_ecc_optimized)
- torque_mig_timescale_helper (torque_mig_timescale_optimized)
- shock_luminosity_helper (shock_luminosity_opt)
- jet_luminosity_helper (jet_luminosity_opt)
- 
Each new function variant is tested inline through an entire 100-galaxy run, and tested on a 25-galaxy sample of inputs drawn directly from actual simulation runs.

Cumulatively, these changes lead to a substantial speedup, with the most recent benchmark showing a runtime reduction from 124 seconds to 106 seconds.

Further optimizations to star_wind_mass_loss and accrete_star_mass are under testing and review, expected to save an additional 9 seconds per run.

In addition, optimizations to analytical_kick_velocity and setup_disk_blackholes_location_NSC_powerlaw have been tested inline, but fragile use of rng in existing testing makes it inconvenient to include them before the restructure. These optimizations have been benchmarked to save 13 seconds.


xevra (approved):
These changes  pass my review.

The new functions are tested well, and all the tests pass on my machine.

The plots from the default runs also look fine.

However, there's a couple of things I wish to discuss as a group on Monday (which I do not believe should hold back this PR, though I will refrain from merging it myself in-case you wish to read it and have a different opinion).

1. The way the tests are now, if somebody wants to update the physics in the Python version of one of these functions, but has no way of updating mcfast, the test will fail. I want us to think about how we plan to handle this. I would think uncommenting the Python version and commenting out the version and mcfast test to be the appropriate procedure, but this sounds inelegant and I want to see if anybody has better ideas.

2. Retaining the commented original versions of each function is correct, and thank you for leaving them in. However, it may not be obvious to new members of the team why these comments are important to retain, and somebody may attempt to delete them in another PR as unnecessary. There is a solution (an off switch for mcfast in the configuration which still includes mcfast as a dependency in the pyproject.toml file) I would propose for this, and I want to make a PR of my own to implement this solution, but we should discuss it as a group because I understand ReadInputs is being entirely forlorn in the refactor.

PR #368, opened 2026-02-18
Eliminating redundant unit conversions

nrposner:
Several functions, most notably r_g_from_units, perform unnecessary SI unit conversions on astropy.constants.c and G.

Both these imports are already denominated in SI units (m/s and m^3 / kg s^2 respectively), per the [documentation](https://docs.astropy.org/en/stable/units/ref_api.html#module-astropy.units.si), which can also be confirmed in the console.

```
>>> import astropy.constants as const
>>> const.G
    class 'astropy.constants.codata2018.CODATA2018'> name='Gravitational constant' value=6.6743e-11 uncertainty=1.5e-15 unit='m3 / (kg s2)' reference='CODATA 2018'>
>>> const.c
    class 'astropy.constants.codata2018.CODATA2018'> name='Speed of light in vacuum' value=299792458.0 uncertainty=0.0 unit='m / s' reference='CODATA 2018'>
```

Converting them into the same units does not change them, but does burn cycles. In r_g_units alone, this accounts for nearly a tenth of total simulation runtime, owing to the frequency with which it is called.

We remove these extraneous conversions from several functions in point_masses, including orbital_separation_evolve and orbital_separation_evolve_reverse, which do not seem to be used in the simulation at present, but may be used in the future.

In addition, we replace string-based unit indication with unit literals: for astropy's unit conversion, `.to("m")` is semantically identical to `.to(u.m)`, but the former requires string parsing, which slows down the conversion. Across several functions, this shaves a couple seconds off the simulation.

Cumulatively, these optimizations reduce profiled time from 104 to 89 seconds, an ~15% runtime speed improvement.

This PR does not introduce new automated tests, as these changes survived inline testing and all changes were judged to be self-evidently correct. If this is not the case and testing is required, please advise.

### DIFF
--- a/src/mcfacts/physics/accretion.py
+++ b/src/mcfacts/physics/accretion.py
@@ -167,7 +167,7 @@ def accrete_star_mass(disk_star_pro_masses,
     timestep_duration_yr_si = timestep_duration_yr * u.year
 
     # Calculate Bondi and Hill radii
-    r_bondi = (2 * const.G.to("m^3 / kg s^2") * star_masses_si / (disk_sound_speed_si ** 2)).to("meter")
+    r_bondi = (2 * const.G * star_masses_si / (disk_sound_speed_si ** 2)).to(u.m)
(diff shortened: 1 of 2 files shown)


xevra (approved):
All tests pass both locally and in the CI.

PR #370, opened 2026-03-05
Optimizing kick_velocity and NSC_powerlaw

nrposner:
Following PR #367, we submit optimized variants of two additional functions: `setup_disk_blackholes_location_NSC_powerlaw`, and `analytical_kick_velocity`. 

Both functions now benefit from mcfast-based helpers to speed up their functionality. In local testing, simulation time went from 49 seconds to 36 seconds, a 35% speedup overall.

- `analytical_kick_velocity` sees an overall speedup of over 200x, with total time spent over the course of a simulation going from 2.8 seconds to 13 milliseconds.
- `setup_disk_blackholes_location_NSC_powerlaw` sees an overall speedup of ~3x, with total time spent over the course of the simulation going from 13 seconds to 3.9 seconds. Further optimizations here are possible, but would be relatively small (at most 2x further with current methods) and would rely on the very small range of current inputs to the function to precompute relevant outputs and introduce a jump table, with the new optimized function as a fallback.

New tests are introduced to check the new and old functions for identical outputs to within rtol 1e-9.

A note: both these tests are placed at the end of the alphabetical pytest ordering in order to prevent them from interfering with test_setup_diskblackholes.py. When either of these run before this other script, it causes this other script to fail, because they too use rng, and the setup test is very fragile. Attempts to fix this issue have been made, and failed. It is not clear why the similar tests introduced recently for accretion do not have the same problem.

We suggest making test_setup_diskblackholes.py more robust and not dependent on global rng state.


xevra (approved):
Passes all tests locally and in CI

`make plots` time 30.96 seconds on Pharaon.

Image

Image

Image

Image

Image

Image

Image

PR #371, opened 2026-03-19
Restructure 

JakePost:
### Summary of Changes

This branch refactors McFACTS, moving away from a monolithic script towards a modular simulation framework. The physics remains, in principal, unchanged. 

The main script driving the simulation `scripts/mcfacts_sim.py` had grown to over 3000 lines, and encapsulated the entire lifecycle of the simulation, including input parsing, simulation setup, the main simulation loop, all of the physics calls, and output writing.

#### Simulation Architecture

In the overhaul of the simulation architecture, a galaxy is assembled by `populators` and evolved through an ordered list `timeline actors`. The `SimulationTimeline` and `TimelineActor` objects in [timeline.py](src/mcfacts/objects/timeline.py) hold the sequence of events that occur throughout a timestep. Unique timelines, handling different physics, can be created for different phases of the simulation. The `Galaxy` object in [galaxy.py](src/mcfacts/objects/galaxy.py) represents a single galaxy and runs timelines against its object population. `AGNObjectArray` and the `FilingCainet` objects in [agn_object_array.py](src/mcfacts/objects/agn_object_array.py) provide access to methods for handling parallel arrays representing populations and objects in the simulation. All objects carry a UUID, generated when the object gets added to the population set, moving towards functions that address objects by id instead of index. 

The `SettingsManager` in [settings_manager.py](src/mcfacts/inputs/settings_manager.py) is the one-stop-shop for declaring the name, category, type, and default value of inputs for the simulation. Instead of receiving specific settings via arguments, Actors access a reference to the main SettingsManager object passed from the parent Galaxy.  With the exception of the `StaticSettingsProperty` type, all properties can be adjusted via an input file, through code (as in a notebook deployment), or through CLI flags. The order of precedence for properties goes as `code-first defaults < input file < CLI flags / notebook initialization). Implementations of `SnapshotHandler` from [snapshot.py](src/mcfacts/objects/snapshot.py) handle the loading and saving of settings, along with the output of intermediate and final simulation states and populations.

#### Command Line Interface, Notebook Support, and Logging

Previously, a Make file was used and maintained as the primary launch point for a general run of McFACTS. In the restructure a console script `mcfacts` ([mcfacts.py](src/mcfacts/mcfacts.py)) exists with three subcommands: `run`, `plot`, and `rp` (run and plot). Settings can be directly passed as flags for the command, and the `--profile` flag wraps the entire run in cProfile.

The `LogFunction` object, and implementations in [log.py](src/mcfacts/objects/log.py), replace the direct calls to the `print` function throughout the simulation. Prefixes containing the galaxy, and the actor now nest automatically allowing logs to be traced back easily. Logging to the console is turned off by default, but can be globally toggled through the use of the `--verbose` flag.

#### General Reorganization

The directory `src/mcfacts/physics` has been removed, with all of the physics methods being moved to `src/mcfacts/modules`, and non-physics / helper methods moved to `src/mcfacts/utilities` (`unit_conversion`, `constants`, `peters`, `random_state`, `checks`). Creating this separation prevents accidental circular imports and provides a natural place for "public" and "private" functions to live.

The default plots produced by [population_plots.py](scripts/population_plots.py) have been reimplemented into [fiducial_plots.py](src/mcfacts/fiducial_plots.py), callable via the CLI and implementing the shared matplotlib style in `src/mcfacts/vis/`.

#### File moved and/or renamed 
_These includes internal reorganizations and function relocation, so files are not one-to-one_

| Old | New |
|---|---|
| `src/mcfacts/physics/accretion.py` | `src/mcfacts/modules/accretion.py` |
| `src/mcfacts/physics/disk_capture.py` | `src/mcfacts/modules/disk_capture.py` |
| `src/mcfacts/physics/disk_capture_stars.py` | `src/mcfacts/modules/disk_capture_stars.py` |
| `src/mcfacts/physics/dynamics.py` | `src/mcfacts/modules/dynamics.py` |
| `src/mcfacts/physics/migration.py` | `src/mcfacts/modules/migration.py` |
| `src/mcfacts/physics/star_interactions.py` | `src/mcfacts/modules/star_interactions.py` |
| `src/mcfacts/physics/stellar_interpolation.py` | `src/mcfacts/modules/stellar_interpolation.py` |
| `src/mcfacts/physics/tde.py` | `src/mcfacts/modules/tde.py` |
| `src/mcfacts/physics/binary/formation.py` | `src/mcfacts/modules/formation.py` |
| `src/mcfacts/physics/binary/merge.py` | `src/mcfacts/modules/merge.py` |
| `src/mcfacts/physics/eccentricity.py` | `src/mcfacts/modules/damping.py` |
| `src/mcfacts/physics/gw.py` | `src/mcfacts/utilities/peters.py` |
| `src/mcfacts/physics/point_masses.py` | `src/mcfacts/utilities/unit_conversion.py` |
| `src/mcfacts/constants.py` | `src/mcfacts/utilities/constants.py` |
| `scripts/em_plots.py` | `scripts/emily_plots.py` |

#### Files absorbed or split

| Old | New |
|---|---|
| `src/mcfacts/physics/feedback.py` | `src/mcfacts/modules/migration.py` |
| `src/mcfacts/physics/lum.py` | `src/mcfacts/modules/merge.py` |
| `src/mcfacts/physics/emri.py` | `src/mcfacts/modules/gw.py` |
| `src/mcfacts/physics/analytical_velocity.py` | `src/mcfacts/modules/merge.py` |
| `src/mcfacts/physics/binary/evolve.py` | `src/mcfacts/modules/accretion.py`, `src/mcfacts/modules/dynamics.py`, `src/mcfacts/modules/gas_hardening.py`, `src/mcfacts/utilities/checks.py` |

#### Files deleted

| File |
|---|
| `src/mcfacts/mcfacts_random_state.py` |
| `src/mcfacts/physics/disk_capture.py.old` |
| `src/mcfacts/physics/__init__.py` |
| `src/mcfacts/physics/binary/__init__.py` |

#### Physics added
- Semi-analytical gas drag, `stalling_separation` and `gas_hardening_prescription` in [gas_hardening.py](ssrc/mcfacts/modules/gas_hardening.py).
- Optional Bondi accretion, `flag_enable_bondi` and `bondi_fraction` in [accretion.py](ssrc/mcfacts/modules/accretion.py)

### Validation Steps

```bash
git clone https://github.com/McFACTS/McFACTS.git
cd McFACTS && git checkout restructure && git merge origin/main-dev
conda create -n my-brand-new-mcfacts-environment python=3.13
conda activate  my-brand-new-mcfacts-environment
conda install numpy==2.3.5 pytest
pip install .
pytest
mcfacts rp
```

### Checklist

- [x] I have checked that no open Pull Requests exist for the same change
- [x] Commit messages follow the Conventional Commits guidelines
- [ ] ~Tests for new core functionalities have been added~ To be completed after merge with `main-dev`
- [ ] ~Added / updated documentation~ To be completed after merge with `main-dev`
- [x] I have added the appropriate label to this Pull Request


xevra:
In a fresh conda environment, the `make plots` (following `pip install --editable .`) fails because `typing_extensions` is not available. Please add this to the pyproject.toml file.

In addition, in dynamics we have:
```
from h2.settings import Settings
```
This is not used after being imported, so rather than added to the pyproject.toml, I believe this line should be commented.

I commented it out on my local branch, and had no trouble running `make plots` without installing `h2`.

scripts/tests/tests.py
```diff
@@ -14,9 +12,9 @@ def test_merger():
     test_bh_angle1 = 1.80
     test_bh_angle2 = 0.7
     test_bin_bh_ang_mom = 1.0
-    test_bh_merg_mass = mcfacts.physics.binary.merge.merged_mass(test_bh_mass1, test_bh_mass2, test_bh_spin1, test_bh_spin2)
```
xevra, line 17: These should be `mcfacts.modules.merge`

scripts/vera_mstar_bins.py
xevra, line 1: Yeah, `point_masses` isn't a file anymore, but this script was never going to survive the restructure. I'll write a replacement on the other side of the review.

src/mcfacts/external/DiskModelsPAGN.py
```diff
@@ -8,10 +8,13 @@
 from pagn import Sirko
 import scipy.interpolate
 
-from mcfacts.physics import point_masses
+import mcfacts.utilities.unit_conversion
```
xevra: This module is imported twice. This is the one which isn't used, so it should be removed.

xevra:
> In a fresh conda environment, the `make plots` (following `pip install --editable .`) fails because `typing_extensions` is not available. Please add this to the pyproject.toml file.
> 
> In addition, in dynamics we have:
> 
> ```
> from h2.settings import Settings
> ```
> 
> This is not used after being imported, so rather than added to the pyproject.toml, I believe this line should be commented.
> 
> I commented it out on my local branch, and had no trouble running `make plots` without installing `h2`.

We're also going to need tqdm.

src/mcfacts/inputs/settings_manager.py
```diff
@@ -0,0 +1,255 @@
...
+    "flag_use_pagn": True,  # Enable or disable the pAGN program
+    "disk_model_name": "sirko_goodman",  # Disk model options: sirko_goodman, thompson_etal
+    "disk_radius_trap": 700.0,  # Migration trap radius (gravitational radii)
+    "disk_radius_outer": 50000.0,  # Outer radius of the disk (gravitational radii)
+    "disk_alpha_viscosity": 0.01,  # Disk viscosity parameter
+    "disk_aspect_ratio_avg": 0.03,  # Average disk height relative to its radius
+    "disk_bh_torque_condition": 0.1,  # Fraction of mass accreted before spin alignment
+    "disk_bh_eddington_ratio": 1.0,  # Eddington accretion ratio
+    "disk_bh_orb_ecc_max_init": 0.9,  # Initial maximum orbital eccentricity
+    "disk_radius_capture_outer": 2.e3,  # Outer radius for capture (gravitational radii)
```
xevra: So, this is in the settings manager, and in the settings.txt file.

You have `disk_radius_capture_outer` set to 2000, and in model_choice.ini we have set it to 1000. Is this intended?

src/mcfacts/inputs/settings_manager.py
```diff
@@ -0,0 +1,255 @@
...
+    "disk_bh_pro_orb_ecc_crit": 0.01,  # Critical eccentricity for circularized orbits
+    "inner_disk_outer_radius": 50.0,  # Outer radius of the inner disk (gravitational radii)
+    "disk_radius_max_pc": 0.,  # Maximum disk size in parsecs (negative for off)
+    "disk_inner_stable_circ_orb": 6.0,  # Innermost stable circular orbit (gravitational radii)
+    "torque_prescription": "paardekooper",  # Torque prescription ('old','paardekooper','jimenez_masset': Paaardekooper default; Jimenez-Masset option)
+    "flag_phenom_turb": False,  # Phenomenological turbulence
+    "phenom_turb_centroid": 0.,  # Centroid of Gaussian w.r.t. to migrating BH. (default = 0.)
+    "phenom_turb_std_dev": 1.0,  # Variance of Gaussian around Centroid (default=0.1)
```
xevra: Here again, this is 1.0, but in model_choice.ini it is 0.1.

The comment indicates that 0.1 is the default, so either the value or the comment should be changed.

src/mcfacts/inputs/settings_manager.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # Black Hole Parameters
+    "mass_pile_up": 35.0,  # Mass pile-up term (solar masses)
+    "initial_binary_orbital_ecc": 0.01, # The initial common orbital eccentricity around the SMBH to be assumed of a binary when it forms
+    "fraction_bin_retro": 0, # Fraction of formed binaries that should be retrograde orbiters
```
xevra: I see that here and in settings.txt it's an int, but it should really be a float.

src/mcfacts/inputs/settings_manager.py
```diff
@@ -0,0 +1,255 @@
...
+    "nsc_imf_bh_powerlaw_index": 2.0,  # Index of the BH mass initial distribution
+    "nsc_imf_bh_mass_max": 40.0,  # Maximum BH mass (solar masses)
+    "nsc_bh_spin_dist_mu": 0.0,  # Mean of the BH spin magnitude distribution
+    "nsc_bh_spin_dist_sigma": 0.1,  # Standard deviation of the BH spin magnitude distribution
+    "nsc_spheroid_normalization": 1.0,  # Sphericity normalization factor
+    "nsc_star_spin_dist_mu": 100,
+    "nsc_star_spin_dist_sigma": 20,
+    "nsc_star_metallicity_x_init": 0.7274,
+    "nsc_star_metallicity_y_init": 0.2638,
+    "nsc_star_metallicity_z_init": 0.0088,
```
xevra: This is also different from model_choice.ini (which has a value of 0.02). Is this intentional?

src/mcfacts/inputs/settings_manager.py
```diff
@@ -0,0 +1,255 @@
...
+    def __init__(self, settings_overrides: dict[str, Any] = None):
+        """
+        Initializes the SettingsManager with a dictionary of overrides.
+
+        Parameters
+        ----------
+        settings_overrides : dict[str, Any]
+            A dictionary containing user-provided overrides for specific settings.
+        """
+        # TODO: Allow users to pass custom settings that we don't have defaults for.
```
xevra: I think this TODO should be removed because I don't think users should be able to add settings to mcfacts without specifying a default value.

JakePost: We can perhaps re-work it by adding a method that creates a new instance of a setting? I was thinking about if someone creates a third-party module that needs its own settings, how would those settings get passed and saved.

Can take a closer look at it in implementing the config parser. Maybe we have a specific section for external settings that third-party modules can use? Code-wise you can add to it to the settings manager, and when you save and load a settings file, it will check for the existence of anything in that custom section? 

We could also just allow items to be added to the settings manager, but not handle their saving. So when someone with a third-party module spins up a run, they'll have to do the loading for custom settings themselves? 

Just some thoughts, we can discuss :)

JakePost: Maybe we need a more robust module declaration system for that though? Either way, there are no pressing things that need it now, I don't think, so its not a hold up for restructure in general.

JakePost: Implemented in https://github.com/McFACTS/McFACTS/pull/377

xevra:
I can't find (either in simulation.py or the tests) an example of a LogFunction being passed to a timeline actor.

Since by default (when parent_log_function is None, the default value for parent_log_function), the timeline actors print *instead* of calling a default LogFunction which prints, it appears that all of the LogFunction support in timeline.py is unreachable.

It might seem pedantic, but you should define a PrintLogFunction with a `__call__` method that prints and use that by default.

This will prevent the LogFunction support from being unreachable, dead code, and will also provide an example for users to write their own LogFunctions with.

Edit: A SimulationTimeline method which updates the log functions of all of its actors would also be handy.

recipes/settings.ini
```diff
@@ -0,0 +1,109 @@
...
+flag_use_pagn = True
+disk_model_name = sirko_goodman
+disk_radius_trap = 700.0
+disk_radius_outer = 50000.0
+disk_alpha_viscosity = 0.01
+disk_aspect_ratio_avg = 0.03
+disk_bh_torque_condition = 0.1
+disk_bh_eddington_ratio = 1.0
+disk_bh_orb_ecc_max_init = 0.9
+disk_radius_capture_outer = 2000.0
```
xevra, line 29: This is different than the default value 1000, which is okay unless this is the fiducial test.

recipes/settings.ini
```diff
@@ -0,0 +1,109 @@
...
+nsc_imf_star_powerlaw_index = 2.35
+nsc_imf_bh_mode = 10.0
+nsc_imf_bh_powerlaw_index = 2.0
+nsc_imf_bh_mass_max = 40.0
+nsc_bh_spin_dist_mu = 0.0
+nsc_bh_spin_dist_sigma = 0.1
+nsc_spheroid_normalization = 1.0
+nsc_star_spin_dist_mu = 100.0
+nsc_star_spin_dist_sigma = 20.0
+nsc_star_metallicity_x_init = 0.7274
```
xevra, line 88: These metallicities also differ from the standard metallicity assumptions we make.

src/mcfacts/objects/actors/reclassify_objects.py
```diff
@@ -0,0 +1,274 @@
...
+
+        # Add prograde disk black holes to separate filing cabinet array.
+        filing_cabinet.create_or_append_array(sm.bh_prograde_array_name, prograde_blackholes)
+
+        # Retrograde black holes
+
+        retrograde_blackholes = blackholes.copy()
+
+        # Find retrograde black holes
+        retrograde_bh_ids = blackholes.unique_id[blackholes.orb_ang_mom < 0]
```
xevra, line 110: There shouldn't ever be any blackholes with `orb_ang_mom == 0`, because they would plunge into the SMBH. If a black hole with `orb_ang_mom == 0`, it's probably an error.

However, if such a black hole was in here somewhere, the reclassification might cause additional errors.

My suggested solution is to take the size of the mask arrays, and sum all three to make sure they add up to the number of black holes that were in here to begin with.

src/mcfacts/objects/actors/reclassify_objects.py
```diff
@@ -0,0 +1,274 @@
...
+
+        self.log("Objects Reclassified")
+
+        # At the end of this, we should have no black holes in the sm.bh_array_name array,
+        # it can be used in the future to seed new black holes before classification is run.
+
+        # TODO: Stars
+
+
+class InnerDiskFilter(TimelineActor):
```
xevra, line 127: There's no docstring here. I know in a 20k PR, linting isn't the top priority, but at least a sentence or two about the purpose of this object would go a long way.

src/mcfacts/objects/actors/reclassify_objects.py
```diff
@@ -0,0 +1,274 @@
...
+        if sm.bh_retrograde_array_name not in filing_cabinet:
+            return
+
+        if sm.bh_prograde_array_name not in filing_cabinet:
+            return
+
+        blackholes_pro = filing_cabinet.get_array(sm.bh_prograde_array_name, AGNBlackHoleArray)
+        blackholes_retro = filing_cabinet.get_array(sm.bh_retrograde_array_name, AGNBlackHoleArray)
+
+        inc_threshhold = 5.0 * np.pi / 180.0
```
xevra, line 234: This is a magic number. Please add a comment, even if it's only "For consistency with mcfacts_sim.py"

src/mcfacts/objects/actors/reclassify_objects.py
```diff
@@ -0,0 +1,274 @@
...
+                agn_disk.disk_aspect_ratio,
+                random_generator
+            )
+
+            stars_to_bh = stars.copy()
+            stars_to_bh.keep_only(star_to_bh_id_num)
+
+            new_blackholes = AGNBlackHoleArray(
+                unique_id= stars_to_bh.unique_id,
+                mass = stars_to_bh.mass,
```
xevra, line 313: More of a physics question. Are we assuming that all black holes which originated as stars lose zero mass in the explosion / direct collapse?

src/mcfacts/utilities/checks.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+
+
+def generate_truncated_normal(mean=0, std=1, lower=0.75, upper=0.85, size=10):
+    a = (lower - mean) / std
+    b = (upper - mean) / std
+    return truncnorm.rvs(a, b, loc=mean, scale=std, size=size)
+
+
+def spin_check(gen_1, gen_2, spin_merged):
```
xevra, line 173: I'm marking this because it's not enabled by default and I don't know in what situations we would want to use it.

This review is not looking at this function for correctness.

src/mcfacts/modules/disk_capture.py
```diff
@@ -315,52 +326,99 @@ def retro_bh_orb_disk_evolve(smbh_mass, disk_bh_retro_masses, disk_bh_retro_orbs
...
     inc_scale_factor[cos_pm1_mask & max_ecc_mask] = step2_time * tau_inc_div[cos_pm1_mask & max_ecc_mask]
-    inc_scale_factor[cos_pm1_mask & barely_prograde_mask] = step3_time * tau_inc_div[cos_pm1_mask & barely_prograde_mask]
+    inc_scale_factor[cos_pm1_mask & barely_prograde_mask] = step3_time * tau_inc_div[
+        cos_pm1_mask & barely_prograde_mask]
     inc_scale_factor[cos_0_mask] = stepw0_time * tau_inc_div[cos_0_mask]
 
     # Calculate new orb_ecc values
-    disk_bh_retro_orbs_ecc_new[cos_pm1_mask & no_max_ecc_retro_mask] = disk_bh_retro_orbs_ecc[cos_pm1_mask & no_max_ecc_retro_mask] * (
-        1.0 + step1_delta_ecc / disk_bh_retro_orbs_ecc[cos_pm1_mask & no_max_ecc_retro_mask] * (timestep_duration_yr / ecc_scale_factor[cos_pm1_mask & no_max_ecc_retro_mask]))
+    disk_bh_retro_orbs_ecc_new[cos_pm1_mask & no_max_ecc_retro_mask] = disk_bh_retro_orbs_ecc[
```
xevra, line 353: I acknowledge that you did not create this issue or change the order of operations, but I'm flagging it because I don't know when the next person will look at it.

Long chains of operators like this without parens worry me.

Best case scenario is that we're relying on Python's order of operations being correct in a way that might confused developers who didn't write this.

Worst case scenario there's a bug here that is also in main-dev.

There also aren't any comments indicating what the order of operations should be, or referencing this equation from somewhere else.

For now, let's at least create a comment to the effect of "TODO: check this again later."

src/mcfacts/simulation.py
```diff
@@ -0,0 +1,193 @@
...
+                                               timesteps=settings.active_timestep_num,
+                                               timestep_length=galaxy.settings.active_timestep_duration_yr)
+
+        # Initial check to make sure our single black holes are real
+        active_phase_timeline.add_timeline_actor(SingleBlackHoleRealityCheck())
+
+        # Get names of different singleton arrays we run through the same module
+        prograde_array = galaxy.settings.bh_prograde_array_name
+        innerdisk_array = galaxy.settings.bh_inner_disk_array_name
+        inner_gw_only_array = galaxy.settings.bh_inner_gw_array_name
```
xevra, line 92: It will be worth noting that we don't yet support retrograde black holes.

src/mcfacts/modules/formation.py
```diff
@@ -327,3 +335,204 @@ def add_to_binary_obj(blackholes_binary, blackholes_pro, bh_pro_id_num_binary, i
...
+            spin=blackholes_pro.get_attribute("spin", primary_ids),
+            spin_2=blackholes_pro.get_attribute("spin", secondary_ids),
+            spin_angle=blackholes_pro.get_attribute("spin_angle", primary_ids),
+            spin_angle_2=blackholes_pro.get_attribute("spin_angle", secondary_ids),
+            bin_sep=bin_sep,
+            bin_orb_a=bin_orb_a,
+            time_to_merger_gw=np.zeros(primary_ids.size),
+            flag_merging=np.zeros(primary_ids.size, dtype=np.int_),
+            time_merged=np.zeros(primary_ids.size, dtype=np.float64),
+            bin_ecc=np.array([random_generator.uniform() for _ in range(primary_ids.size)], dtype=np.float64),
```
xevra, line 522: This appears uniform between 0 and 1. Is that intended?

src/mcfacts/modules/gw.py
```diff
@@ -0,0 +1,428 @@
...
+    return new_disk_bh_pro_orbs_a
+
+
+def gw_hardening(mass_1, mass_2, bin_ecc, bin_sep, bin_time_to_merge, flag_merging, smbh_mass, timestep_length, r_g_in_meters):
+    array_length = len(mass_1)
+
+    flag_not_merging = np.array([flag_merging[i] >= 0 for i in range(array_length)], dtype=np.bool_)
+
+    # Find eccentricity factor (1-e_b^2)^7/2
+    ecc_factor_1 = np.power(1 - np.power(bin_ecc[flag_not_merging], 2), 3.5)
```
xevra, line 282: Note to self for next part of the review. Check to. see if ** is faster than np.power
edit: disregard

src/mcfacts/modules/migration.py
```diff
@@ -926,3 +985,586 @@ def type1_migration_binary(smbh_mass, bin_mass_1, bin_mass_2, bin_orb_a, bin_orb
...
+
+                if sm.flag_thermal_feedback == 1:
+                    total_jimenez_masset_torque_bh = jimenez_masset_torque_coeff_bh + jimenez_masset_thermal_torque_coeff_bh
+                else:
+                    total_jimenez_masset_torque_bh = jimenez_masset_torque_coeff_bh
+
+                torque_bh = total_jimenez_masset_torque_bh * normalized_torque_bh
+
+                # Set up trap scaling as a function of mass for Jimenez-Masset (for SG-like disk)
+                # No traps if M_smbh >10^8Msun (approx.)
```
xevra, line 1341: This is new, and I wasn't expecting it. Why are we not scaling to higher masses?

src/mcfacts/modules/gas_hardening.py
xevra, line 1: This file is not commented. I'm not reviewing it for physical accuracy at this time.

src/mcfacts/modules/accretion.py
xevra, line 1: I'm not reviewing this for physical accuracy at this time.

notebooks/stahler_test.ipynb
xevra, line 1: Last cell struggles with new settings.

notebooks/galaxy_tester.ipynb
xevra, line 1: Running into issues running this from jupyter-lab, because scripts is a folder, and the notebooks are not in the same folder. At some point, it might be worth reorganizing a few things in scripts so they can be imported from the module. That is not a requirement of the review.

notebooks/ecc_v_time.ipynb
xevra, line 1: This isn't working for me when I run `mcfacts rp -o true`. However, the notebooks working is not a requirement for this merge.

notebooks/bondi_tester.ipynb
xevra, line 1: Call signatures have changed since this was written. Right now it barfs.

xevra:
I am investigating a bug related to custom settings files that will need to be fixed before approval can be given, but I haven't nailed it down yet, so the rest of the comments are still worth responding to in the meantime.

edit: The bug was mine.

src/mcfacts/simulation.py
```diff
@@ -0,0 +1,193 @@
...
+        assert False, f"Output directory {settings.output_dir} already exist. Set --overwrite_files=True to clear the directory."
+
+    if settings.overwrite_files and os.path.isdir(settings.output_dir):
+        shutil.rmtree(settings.output_dir)
+
+    # Create the IO handlers and save the current settings
+    snapshot_handler = TxtSnapshotHandler(settings = settings)
+
+    ini_handler = IniSnapshotHandler(settings=settings)
+    ini_handler.save_settings("./runs", "settings", settings)
```
xevra: This is hard coded as runs, instead of the user specified output directory.

src/mcfacts/simulation.py
```diff
@@ -0,0 +1,193 @@
...
+    for galaxy_id in range(settings.galaxy_num):
+        pbar.set_description(f"Running Galaxy {galaxy_id}")
+        pbar.update(1)
+
+        # The Galaxy class creates a random generated based on this seed,
+        # Philox output for n and n+-1 seeds are uncorrelated
+        galaxy_seed = settings.seed - galaxy_id
+
+        # Create instance of galaxy
+        galaxy = Galaxy(seed=galaxy_seed, runs_folder="./runs", galaxy_id=str(galaxy_id), settings=settings)
```
xevra: This too.

src/mcfacts/simulation.py
```diff
@@ -0,0 +1,193 @@
...
+        if inner_gw_only_array in galaxy.filing_cabinet:
+            population_cabinet.create_or_append_array("blackholes_emri", galaxy.filing_cabinet.get_array(inner_gw_only_array))
+
+        if emri_merged_array in galaxy.filing_cabinet:
+            population_cabinet.create_or_append_array("blackholes_emri", galaxy.filing_cabinet.get_array(emri_merged_array))
+
+    pbar.close()
+
+    # Save the entire population cabinet
+    snapshot_handler.save_cabinet("./runs", "population", population_cabinet)
```
xevra: This is also hard-coded as runs.

xevra (changes requested):

src/mcfacts/mcfacts.py
```diff
@@ -0,0 +1,145 @@
...
+        loaded_settings = initial_settings
+
+    static_settings = [prop.name for prop in settings_manager.DEFAULT_SETTINGS if isinstance(prop, StaticSettingsProperty)]
+
+    # Parse through the loaded settings and create the corresponding arguments with the loaded settings as defaults
+    for key, value in loaded_settings.settings_finals.items():
+        if key in static_settings:
+            continue
+
+        if key is "settings_file":
```
xevra: Bad use of is. Use == for string comparison.

xevra:
Also, pytest doesn't work, but I imagine that's the last thing you'll want to fix, because changing things might break it again.

recipes/settings.ini
JakePost: Settings manager changes revert this back to 1000.

recipes/settings.ini
JakePost: https://github.com/McFACTS/McFACTS/pull/371#discussion_r3713223979

src/mcfacts/objects/actors/reclassify_objects.py
JakePost: Added docstrings in [cec7f64](https://github.com/McFACTS/McFACTS/pull/371/commits/cec7f6477fa53cad80a407d9ee8bbbd0690a38d3)

src/mcfacts/objects/actors/reclassify_objects.py
JakePost: Added comment in [cec7f64](https://github.com/McFACTS/McFACTS/pull/371/commits/cec7f6477fa53cad80a407d9ee8bbbd0690a38d3), and todo for moving it to settings manager.

src/mcfacts/objects/actors/reclassify_objects.py
JakePost: Definitely a question for @spacekaila

src/mcfacts/utilities/checks.py
JakePost: Question for @shray4

src/mcfacts/modules/disk_capture.py
JakePost: Agreed, added a TODO comment, we can reexamine later, and maybe breakout the inline operations.

src/mcfacts/simulation.py
JakePost: Added a comment. We do have an actor that handles retrograde evolution, but we don't apply the same treatment to retrogrades as we do with other single objects. Something to consider expanding with @saavikford

src/mcfacts/modules/formation.py
JakePost: Currently it maintains parity with the "add to binary" code that was used in mcfacts_sim.py:

Line 241 in formation.py
```
# Draw uniform value btwn [0,1]
bin_ecc = random.uniform(size=bin_num)
```

src/mcfacts/modules/gw.py
JakePost: Resolved?

src/mcfacts/modules/migration.py
JakePost: IIRC Resolved during group meeting? Question for @bmckernan if not.

src/mcfacts/modules/gas_hardening.py
JakePost: Added docstrings and some comments throughout.

notebooks/stahler_test.ipynb
JakePost: Removed the stalling-paper specific files. Will move them into their own repo or branch.

notebooks/galaxy_tester.ipynb
JakePost: The notebooks should be working with McFACTS as an installed package, so it shouldn't need to point to the internal script folders. The tester files mostly exist as examples.

notebooks/ecc_v_time.ipynb
JakePost: This was a small test I ran to debug stuff early in restruct dev. Looks like the file names it expects are different. I agree, we can fix this in the future.

src/mcfacts/simulation.py
JakePost: [9a025a7](https://github.com/McFACTS/McFACTS/pull/371/commits/9a025a77e0cfc7cff1cc15664f5a984fd83deda1) Updates this to now use the output dir defined in the settings.

src/mcfacts/simulation.py
JakePost: https://github.com/McFACTS/McFACTS/pull/371#discussion_r3714769377

src/mcfacts/simulation.py
JakePost: https://github.com/McFACTS/McFACTS/pull/371#discussion_r3714769377

src/mcfacts/simulation.py
JakePost: Also updated the hardcoded "./runs" that appears in fiducial_plots.py

src/mcfacts/mcfacts.py
JakePost: ooo, good catch! Fixed in [4a1fe4f](https://github.com/McFACTS/McFACTS/pull/371/commits/4a1fe4fd3643df44012d0939bc5d7e2fd8a38e50)

notebooks/bondi_tester.ipynb
JakePost: Will be moving this to a separate repo or branch when work on bondi starts up again.

src/mcfacts/modules/migration.py
xevra: I think the decision we came to is that it should only be enforced for the jimenez_masset_torque model instead of universally (what it is now).

recipes/settings.ini
xevra: Broken link?

recipes/settings.ini
xevra: Do you mean that the settings manager overrides the inifile?

If we load the inifile, the inifile should take priority over the hard-coded defaults.

Is settings.ini intended to represent a fiducial run?

src/mcfacts/objects/actors/reclassify_objects.py
xevra: I still think a check here could save us a lot of trouble later.

src/mcfacts/utilities/checks.py
xevra: I don't think this should hold back this PR.

src/mcfacts/objects/actors/reclassify_objects.py
xevra: I don't think this should hold back this PR, though whoever is working on stars should come back and look at this.

src/mcfacts/external/DiskModelsPAGN.py
xevra: Unresolved

src/mcfacts/modules/formation.py
xevra: That's not great, and we should raise it to the group, but isn't an issue with restructure.

xevra:
Five unresolved conversations remain (one of them might be hard to find; extra import commented on May 18th).

Pytest passes with scipy < 1.18 on my machine. Passes in PR CI with changes from main-dev. I'm old fashioned, so I like to see changes from main-dev merged back into the branch, but this is optional.

output_dir parameter works as intended (tested locally).

src/mcfacts/objects/actors/reclassify_objects.py
JakePost: Per mcfacts_sim.py, line 494,

`# Find prograde BH orbiters. Identify BH with orb. ang mom > 0 (orb_ang_mom is only ever +1 or -1)`

Maybe this is incorrect now? But the current functionality copies the functionality of mcfacts_sim.py -- I'd say we should reconsider after the PR is resolved if we need to.

src/mcfacts/external/DiskModelsPAGN.py
JakePost: Removed double import.

src/mcfacts/modules/formation.py
JakePost: Per Saavik & Barry: Update so that we assume circular for all new binaries. In the future we can implement a thermal distribution, but will need update how we treat binary eccentricity for it to actually matter.

src/mcfacts/modules/migration.py
JakePost: Just checking if the current state of the code is where we want it to be for now, or if something needs to be updated.

xevra:
PR open to resolve LogFunction issues: https://github.com/McFACTS/McFACTS/pull/383

src/mcfacts/modules/migration.py
xevra: It's not. Right now, this happens for any given migration prescription, and was only intended to be enforced for jimenez_masset. 

It would also be better if we could turn it on/off.

Please add a flag to settings for this.

xevra (approved):
At this time, all of my questions and concerns about this pull request have been addressed. I hereby approve the restructure pull request on September 2nd, 2026.

PR #372, opened 2026-04-15
performance: unit conversions, star gating, circular singles

nrposner:
You've been waiting for it: the biggest performance PR to date, focused on eliminating and streamlining unnecessary unit conversions. Headline times: in local testing with profiling, brings time from ~30 seconds to ~18 seconds on a Mac M3 Max.

- Added variant optimized function to replace circular_singles_encounters_prograde and circular_singles_encounters_prograde_sweep with new, faster helper function, circular_singles_encounters_prograde_sweep_optimized. This takes runtime of the function down from ~3 seconds to ~0.12 seconds on local testing.

- Gate off accrete_star_mass and star_wind_mass_loss functions when `stars` are turned off

- Added optimized variant of r_schwarzschild that reduces use of expensive unit conversions. Takes function runtime with profiling from 17 seconds to 0.15 seconds
- - Also condensed existing r_schwarzschild calls, as `r_schwarzschild(a) + r_schwarzschild(b)` into `r_schwarzschild(a+b)`, since the function's transformation is commutative

- Added optimized variants of si_from_r_g and r_g_from_units functions. Take function runtime with profiling from 7.3 and 9.0 seconds to 0.46 and 0.47 seconds respectively

- Eliminated some unit conversions which evaluated to constants, replacing them with explicit constants, such as 
```
# original
solar_mass = u.solMass.to("kg")

# replacement with predefined constant
solar_mass = M_SUN_KG
```

- Added test cases for all new function variants, testing that outputs are identical to within either relative tolerance 1e-6 or 1e-9


xevra:
19.87 seconds on Pharaon
gw_strain
mass_radius
4ab-f0d089cae27f" />
time_of_merger
merger_remnant_mass
merger_mass_v_radius

xevra:
Performance time: 296.63  [ 4 min - 56 sec ]
10 Myr
[PR372.prof.gz](https://github.com/user-attachments/files/26765665/PR372.prof.gz)

scripts/mcfacts_sim.py
```diff
@@ -1108,28 +1112,34 @@ def main():
...
-                        opts.smbh_mass,
-                        blackholes_pro.orb_a,
-                        blackholes_pro.mass,
-                        blackholes_pro.orb_ecc,
-                        opts.timestep_duration_yr,
-                        opts.disk_bh_pro_orb_ecc_crit,
-                        opts.delta_energy_strong_mu,
-                        opts.disk_radius_outer
-                    )
+                # # --- BEGIN: capture sweep inputs for test generation ---
```
xevra: If leaving the comments in is intentional, I think that's fine. If it's intentional.

scripts/mcfacts_sim.py
```diff
@@ -2052,7 +2065,15 @@ def main():
 
                 # Harden binaries via gas
                 # Choose between Baruteau et al. 2011 gas hardening, or gas hardening from LANL simulations. To do: include dynamical hardening/softening from encounters
-                blackholes_binary.bin_sep, blackholes_binary.flag_merging, blackholes_binary.time_merged, blackholes_binary.time_to_merger_gw = evolve.bin_harden_baruteau(
+
```
xevra, line 2050: More comments that are fine if intentional, and if you can explain why we should leave them in.

scripts/mcfacts_sim.py
```diff
@@ -2066,6 +2087,27 @@ def main():
                     time_passed,
                     opts.r_g_in_meters)
 
+
```
xevra: See other comments about comments.

xevra (commented):
Looks good to me. Plots look fine. Code looks fine. Pytest passes locally.

I don't want to approve the PR because I made some comments, and it's after hours, and nobody else has had time to look at it yet.

Conditional approval 1 PM Monday, after the meeting.

scripts/mcfacts_sim.py
nrposner: not intentional, removed

xevra (approved):

PR #377, opened 2026-06-09
Restructure | Implement Config Parser Save/Load

JakePost:
Implemented .ini file handling for settings save/load.

New default setting property declaration (SettingsProperty, StaticSettingsProperty, and OptionalSettingsProperty)

Custom settings from third-party modules can be added code-first and loaded/saved with the ini file.

Changed default settings file from settings.txt to settings.ini. Passing the "--settings_file" argument with a path to a .ini or .txt file will automatically switch to the correct file loader.

### DIFF
--- a/src/mcfacts/inputs/settings_manager.py
+++ b/src/mcfacts/inputs/settings_manager.py
@@ -44,6 +44,11 @@ class SettingsProperty:
         return self._value
 
 
+    @value.setter
+    def value(self, value: T):
+        self._value = value
+
+
     @property
     def prop_type(self) -> Type[T]:
         return self._prop_type
--- a/src/mcfacts/inputs/settings_manager.py
+++ b/src/mcfacts/inputs/settings_manager.py
@@ -94,7 +94,7 @@ class OptionalSettingsProperty(SettingsProperty):
         super().__init__(name, category, value, prop_type)
 
 
-default_settings: list[SettingsProperty | StaticSettingsProperty] = [
+DEFAULT_SETTINGS: list[SettingsProperty | StaticSettingsProperty] = [
         # IO Parameters
         SettingsProperty("verbose", "io", False, bool),
         SettingsProperty("show_timeline_progress", "io", False, bool),
@@ -352,7 +352,7 @@ class SettingsManager:
         self.settings_finals: dict[str, Any] = {}
         category_props: dict[str, dict[str, Any]] = {}
 
-        for prop in default_settings:
+        for prop in DEFAULT_SETTINGS:
             is_static = isinstance(prop, StaticSettingsProperty)
             override = settings_overrides.get(prop.name, prop.value)
             final_value = prop.value if is_static else self._cast_override(prop, override)
--- a/src/mcfacts/inputs/settings_manager.py
+++ b/src/mcfacts/inputs/settings_manager.py
@@ -1,161 +1,343 @@
-from typing import Any
+import warnings
+from functools import cached_property
+from types import NoneType
+from typing import Any, TypeVar, Type
 
 from mcfacts.inputs import ReadInputs
 from mcfacts.utilities import unit_conversion
 
-defaults = {
-    # IO Parameters
-    "verbose": False, # Print all debug messages
-    "show_timeline_progress": False, # Shows a progress bar for the active timeline
-    "overwrite_files": False, # Override any output files that exist, throws errors otherwise
-    "save_state": False, # Pickle and save the entire state of a galaxy after each population or timeline is run
-    "save_each_timestep": False, # Pickle and save the state of a galaxy for each timestep during a simulation timeline
-    "save_snapshots": 0, # :: DEPRECIATED :: Whether to save snapshots (0 for off)
-    "output_dir": "./runs", # Output directory relative to the base directory
-    "settings_file": "./recipes/model_choice_old.ini",
 
-    # Simulation Parameters
-    "active_timestep_duration_yr": 1.e4,  # Duration of each timestep (years)
-    "active_timestep_num": 70,  #  Number of timesteps in dynamics timeline
-    "capture_time_yr": 1.e5,  # Time between disk captures (years)
-    "galaxy_num": 100,  #  Number of iterations of the simulation
-    "seed": 223849053863469657747974663531730220530, # Seed of the simulation, should be 128 bits long
+class SettingsProperty:
+    T = TypeVar("T")
 
-    # AGN Parameters
-    "smbh_mass": 1.e8,  # Supermassive black hole mass (solar masses)
+    def __init__(self, name: str, category: str, value: Any, prop_type: Type[T] = str):
+        assert type(value) == prop_type, f"SettingsProperty: the type of value ({type(value)}) must match prop_type {prop_type}"
 
-    # Disk Parameters
-    "flag_use_pagn": True,  # Enable or disable the pAGN program
-    "disk_model_name": "sirko_goodman",  # Disk model options: sirko_goodman, thompson_etal
-    "disk_radius_trap": 700.0,  # Migration trap radius (gravitational radii)
-    "disk_radius_outer": 50000.0,  # Outer radius of the disk (gravitational radii)
-    "disk_alpha_viscosity": 0.01,  # Disk viscosity parameter
-    "disk_aspect_ratio_avg": 0.03,  # Average disk height relative to its radius
-    "disk_bh_torque_condition": 0.1,  # Fraction of mass accreted before spin alignment
-    "disk_bh_eddington_ratio": 1.0,  # Eddington accretion ratio
-    "disk_bh_orb_ecc_max_init": 0.9,  # Initial maximum orbital eccentricity
-    "disk_radius_capture_outer": 2.e3,  # Outer radius for capture (gravitational radii)
-    "disk_bh_pro_orb_ecc_crit": 0.01,  # Critical eccentricity for circularized orbits
-    "inner_disk_outer_radius": 50.0,  # Outer radius of the inner disk (gravitational radii)
-    "disk_radius_max_pc": 0.,  # Maximum disk size in parsecs (negative for off)
-    "disk_inner_stable_circ_orb": 6.0,  # Innermost stable circular orbit (gravitational radii)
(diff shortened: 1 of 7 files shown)


src/mcfacts/mcfacts.py
```diff
@@ -27,7 +29,7 @@ def seed_settings_args(sub_parser: argparse.ArgumentParser):
     initial_settings = SettingsManager()
 
     # Create a specific flag for passing in a baked settings or ini file
-    sub_parser.add_argument("-s", "--settings", "--fname-ini",
+    sub_parser.add_argument("-s", "--settings", "--fname-ini", "--settings_file",
```
xevra: Is there a reason for not using the dashes like we do elsewhere (I.e. `--settings-file`)?

src/mcfacts/inputs/settings_manager.py
```diff
@@ -1,162 +1,333 @@
...
+
+
+    @category.setter
+    def category(self, value: str):
+        self._category = value
+
+
+    @property
+    def value(self) -> T:
+        return self._value
```
xevra, line 44: It's curious to me that you can change the type and name of a setting after creation, but not its value.

src/mcfacts/inputs/settings_manager.py
```diff
@@ -1,162 +1,333 @@
...
+
+class CategoryProxy:
+    """
+    A proxy object that holds settings for a single property category
+    """
+
+    def __init__(self, props: dict[str, Any] = None):
+        self._props: dict[str, Any] = props if props is not None else {}
+
+    def __getattr__(self, item: str) -> Any:
```
xevra, line 217: If I had written this, I would have done it like:

```
from collections import UserDict
class CategoryProxy(UserDict):

  def __init__(self, props: dict[str, Any] = None, **kwargs):
    self._locked = False
    super().__init__(props, **kwargs)
    self._locked = True

  def __setitem__(self, key: str, value: Any):
     if self._locked:
       raise TypeError("read-only")
     super().__setitem__(key, value)

  def __getitem__(self, key: str) -> Any:
    ...
```
Then you can uses `mycategoryproxy['mysetting']` instead of `mycategoryproxy.get('mysetting')`, and you get things like `__contains__` for free because all of that work was already done for dictionaries, and you can still define custom getters and setters.

I am not asking you to change everything.

I'm just curious why you did it this way instead?

src/mcfacts/inputs/settings_manager.py
```diff
@@ -1,162 +1,333 @@
...
+        if item.startswith("_"):
+            raise AttributeError(f"No private attribute '{item}'")
+        try:
+            return self._props[item]
+        except KeyError:
+            raise AttributeError(f"No setting '{item}' in this category")
+
+    def __repr__(self):
+        return f"CategoryProxy({self._props})"
```
xevra, line 227: `__contains__` would be useful here, so you could write 

```
if "mysetting" in mycategoryproxy:
  do stuff
```

src/mcfacts/inputs/settings_manager.py
```diff
@@ -1,162 +1,333 @@
...
+
+
+class OptionalSettingsProperty(SettingsProperty):
+    T = TypeVar("T")
+
+    def __init__(self, name: str, category: str, value: Any, prop_type: Type[T] = str):
+        super().__init__(name, category, value, prop_type)
+
+
+default_settings: list[SettingsProperty | StaticSettingsProperty] = [
```
xevra: For something like this, all caps is appropriate: `DEFAULT_SETTINGS: ...`

It helps developers to know that this is data that lives in the top level scope of the script.

src/mcfacts/inputs/settings_manager.py
```diff
@@ -165,61 +336,98 @@ def __init__(self, settings_overrides: dict[str, Any] = None):
...
-            if key not in self.settings_overrides:
-                self.settings_overrides[key] = value
+            self.settings_finals[prop.name] = final_value
+            category_props.setdefault(prop.category, {})[prop.name] = final_value
 
-            if (key in self.settings_overrides) and (key not in static_settings):
-                self.settings_finals[key] = self.settings_overrides[key]
-            else:
-                self.settings_finals[key] = value
+        self._categories = {cat: CategoryProxy(props) for cat, props in category_props.items()}
```
xevra, line 374: Before this categories line, it might be worthwhile to add a check here for any settings which are in `settings_overrides`, but which are not in `settings_finals`, and raising some kind of error.

src/mcfacts/objects/snapshot.py
```diff
@@ -209,24 +210,100 @@ def load_settings(self, file_path: str | bytes | PathLike, file_name: str | byte
...
+
+    def save_cabinet(self, file_path: str | bytes | PathLike, file_name: str | bytes | PathLike, filing_cabinet: FilingCabinet):
+        raise NotImplementedError("IniSnapshotHandler does not support saving FilingCabinets")
+
+    def load_cabinet(self, file_path: str | bytes | PathLike, file_name: str | bytes | PathLike) -> Any:
+        raise NotImplementedError("IniSnapshotHandler does not support loading FilingCabinets")
+
+    def save_settings(self, file_path: str | bytes | PathLike, file_name: str | bytes | PathLike,
+                      settings: SettingsManager = None):
+        directory = Path(file_path)
```
xevra: file_path is a confusing name because it is not a file (it should point to a directory) and not a Path (it can be a string).

src/mcfacts/objects/snapshot.py
```diff
@@ -209,24 +210,100 @@ def load_settings(self, file_path: str | bytes | PathLike, file_name: str | byte
...
+        directory.mkdir(parents=True, exist_ok=True)
+
+        final_path = os.path.join(file_path, file_name + ".ini")
```
xevra: I would put this in an else block immediately below the if statement here.

src/mcfacts/objects/snapshot.py
```diff
@@ -209,24 +210,100 @@ def load_settings(self, file_path: str | bytes | PathLike, file_name: str | byte
...
+            for name, value in proxy._props.items():
+                config.set(category, name, str(value))
+
+        with open(final_path, "w") as f:
+            config.write(f)
+
+        self.log(f"Saved settings to {final_path}")
+
+    def load_settings(self, file_path: str | bytes | PathLike, file_name: str | bytes | PathLike) -> SettingsManager:
+        final_path = os.path.join(file_path, file_name + ".ini")
```
xevra: Belongs in an else block after the following if statement.

xevra (changes requested):
Mostly minor suggestions and clarifications. The else block comments really should be addressed though.

src/mcfacts/mcfacts.py
JakePost: Typo on my end! Fixed in [bcba40d](https://github.com/McFACTS/McFACTS/pull/377/commits/bcba40d0a71510e6e1b4c4fff497642bd784aaeb)

src/mcfacts/inputs/settings_manager.py
JakePost: Overlooked this on the first pass, added in [2711deb](https://github.com/McFACTS/McFACTS/pull/377/commits/2711deb202e0c0a9a137c1d3a28878446c4bb783)

src/mcfacts/inputs/settings_manager.py
JakePost: My goal was to copy the functionality of how the SettingsManager object works currently, in that you can reference a setting using `manager.setting` and with a category you can `manager.category.setting`

If we would find it useful to address things like a dictionary, this is probably something we also want to add to SettingsManager, as well as the proxy class. Lmk!

src/mcfacts/inputs/settings_manager.py
JakePost: https://github.com/McFACTS/McFACTS/pull/377#discussion_r3455501828

src/mcfacts/inputs/settings_manager.py
JakePost: Updated in [e787805](https://github.com/McFACTS/McFACTS/pull/377/commits/e78780558f52c17d058a48d1477f781cd8b31327)

src/mcfacts/inputs/settings_manager.py
JakePost: Ooo, good catch! Added a check and a user-level warning if someone passes a custom setting that doesn't exist. [9b8e937](https://github.com/McFACTS/McFACTS/pull/377/commits/9b8e93735ca965a9ffc697e61147efcf82487259)

src/mcfacts/objects/snapshot.py
JakePost: This was probably the hardest change to make because I wasn't sure what else to call it, but I'm definitely overthinking it 😅 

I changed it to "directory" in [85b4697](https://github.com/McFACTS/McFACTS/pull/377/commits/85b469742b4cd20c3580d7789760d608c24dc935), which I think is what it should be? But happy to change it if something else makes more sense!

src/mcfacts/objects/snapshot.py
JakePost: Moved into an else branch [61140db](https://github.com/McFACTS/McFACTS/pull/377/commits/61140db287bdbe8279aebbaaf8280060f7a13030), I always forget you don't need to initialize variables outside of if-else-blocks

src/mcfacts/objects/snapshot.py
JakePost: https://github.com/McFACTS/McFACTS/pull/377#discussion_r3455533287

src/mcfacts/mcfacts.py
JakePost: Oh, I just realized why I used the underscore vs a dash there: It wasn't specifically a conscious choice, but the settings manager has all of its settings defined with underscores instead of dashes.

src/mcfacts/inputs/settings_manager.py
xevra: My sense is that the different approaches may be interesting to explore, but not before restructure is merged.

xevra (approved):
Looks good! Thanks for responding to my feedback!

PR #383, opened 2026-08-07
Restructure logs

xevra:
### Summary of Changes

- Replaced the no-op logging protocol with an enforced object structure.
- Created tests for the new log objects
- Added PrintLogFunction for consistency with current behaviour
- Added ContextLogFunction, with tests to help future support of logging using a context manager.

### Validation Steps

`pip install .; pytest`

Optionally:
```
git checkout restructure
mcfacts run --overwrite_files=True --verbose=True > out1.log
git checkout restructure-logs
mcfacts run --overwrite_files=True --verbose=True > out2.log
diff out1.log out2.log
```

### Checklist

- [] Commit messages follow the Conventional Commits guidelines
    I did my best, but I missed one or two.
- [X] Added / updated documentation

### DIFF
--- a/src/mcfacts/objects/galaxy.py
+++ b/src/mcfacts/objects/galaxy.py
@@ -18,7 +18,9 @@ class GalaxyPopulator(ABC):
     def __init__(self, name: str, settings: SettingsManager = SettingsManager()):
         self.name: str = name
         self.settings: SettingsManager = settings
-        self.parent_log_func: LogFunction = PrintLogFunction
+        self.parent_log_func: LogFunction = PrintLogFunction(
+            preamble=f"(ID:??) {self.name} :: "
+        )
 
     @abstractmethod
     def populate(self, agn_disk: AGNDisk, random_generator: Generator) -> AGNObjectArray:
@@ -35,7 +37,7 @@ class GalaxyPopulator(ABC):
     def log(self, msg: str, new_line: bool = False) -> None:
         if not self.settings.verbose:
             return
-        self.parent_log_func(f"(ID:??) {self.name} :: {msg}")
+        self.parent_log_func(msg)
 
     def __str__(self) -> str:
         """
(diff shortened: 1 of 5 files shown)


src/mcfacts/objects/galaxy.py
```diff
@@ -106,7 +109,10 @@ def __init__(self, seed: int, runs_folder: str, galaxy_id: str, settings: Settin
 
         # Set the recursion limit higher so python doesn't scream at us. The timeline-actor framework does not do any recursion,
         # but when an actor performs, something can end up executing several "layers" away from the initial call.
-        sys.setrecursionlimit(10000)
+        sys.setrecursionlimit(10000) # :'(
```
JakePost, line 112: I also hate having this hard coded. We can definitely take a closer look at this, but running the benchmark and looking at the flame graph doesn't show anything particularly worrying.

src/mcfacts/objects/galaxy.py
```diff
@@ -222,11 +232,13 @@ def run(self, simulation_timeline: SimulationTimeline, agn_disk: AGNDisk) -> Non
...
+        if new_line:
+            self.parent_log_func.new_line()
+        self.parent_log_func(msg)
 
     def log(self, msg: str, new_line: bool = True) -> None:
         if not self.settings.verbose:
             return
-
-        self.nocheck_log(msg, new_line)
+        self.nocheck_log(msg, new_line = new_line)
```
JakePost: The "nocheck_log" function was originally meant to serve as the default log "object" IIRC. But I see no reason not to just collapse this into a single function here.

src/mcfacts/objects/log.py
```diff
@@ -1,28 +1,147 @@
...
 
-    Any callable that matches this protocol should accept a message as a string
-    and an optional `new_line` flag (defaulting to True), and return None.
-    """
+######## Class instances ########
+#### Print Log Function ####
+class PrintLogFunction(LogFunction):
+    """A Log function which calls print"""
+    def __init__(self, preamble=""):
+        self._preamble = preamble
```
JakePost: I think preamble was slightly confusing for me on initial look, since I'm not too used to seeing the word in a coding context -- "prefix" I think is a bit clearer, but its totally ok if it stays "preamble."

JakePost (approved):
Left some comments/suggestions for things that can be changed, but nothing game-breaking and good to merge as is.

Approving, with resolution of comments being optional!

src/mcfacts/objects/log.py
xevra: Done.

src/mcfacts/objects/galaxy.py
xevra: Done.

PR #388, opened 2026-09-03
Scaling

xevra:
This pull request creates a larger, more versatile AGNDisk object, and adds preprocessing options for running McFACTS with an arbitrarily sized supermassive black hole and nuclear star cluster mass.

It replaces the need for the `vera_mstar_runs.py` script.

It does not guarantee that any particular set of simulation parameters will not crash, but it adds the infrastructure required to test differently scaled disks.

scaled_radius_scaling
scaled_nsc_scaling
fixed_radius_scaling
fixed_nsc_scaling

#### New settings
Creates 9 new settings for controlling how the scaling relations used in paper 3 can be used on a simulation with different properties:
- `disk_truncation` : a string that decides which disk truncation method should be used, if any. Default is "none", and I add support for "opacity-equals-inner-disk", which truncates the outer disk where the opacity level matches the opacity level near the black hole. This is the only 'scaling' setting which can be applied when `flag_use_scaling` is False.
- `flag_use_scaling`: A boolean that decides if the rest of the scaling relations are applied. This allows the other settings to have meaningful defaults without enforcing them.
- `stellar_mass`: A float that indicates (in solar mass) what the stellar mass of the AGN's host galaxy should be. WARNING: if `flag_use_scaling` is on and `stellar_mass` is not equivalent to zero, this will override the `smbh_mass` and `nsc_mass` settings using the appropriate scaling relation if not "none". 
- `scale_smbh_mass`: A string choosing a method for estimating the power law relation between supermassive black hole mass and stellar mass of a galaxy, inviting "schramm-silverman" for the [Schramm & Silverman 2013](http://doi.org/10.1088/0004-637X/767/1/13) power law. 'none' otherwise.
- `scale_nsc_mass`: A string choosing a method for estimating the nuclear star cluster mass of a galaxy based on its stellar mass. Valid choices are `["none", "neumayer-early", "neumayer-late"]`. See also [Neumayer+2020](http://doi.org/10.1007/s00159-020-00125-0).
- `scale_inner_disk`: A string indicating a method for scaling the inner disk. Currently the choices are ["none", "decay-time"], which keeps decay time constant for the inner disk, as the SMBH is scaled when set to "decay-time".
- `scale_trap`: A string indicating a method for scaling the trap radius. Currently the choices are ["none", "sqrt-smbh"], where "sqrt-smbh" multiplies `disk_radius_trap` by `np.sqrt(SMBH_MASS_FIDUCIAL / settings.smbh_mass)`.
- `scale_capture_radius`: A string indicating a method for scaling the capture radius. Choices are ["none", "sqrt-smbh"]
- `scale_capture_time`: A string indicating a method for scaling the capture time. Choices are ["none", "hubble", and "WZL+2024"]. A choice of `hubble` effectively turns off dynamic captures. A choice of `WZL+2024` follows a [Wang, Zhu, and Lin (2024)](https://academic.oup.com/mnras/article/528/3/4958/7596571).

#### Modules
- New `set_preprocessing` function in `settings_manager.py`
- `inputs/scaling.py` is where all of the scaling relations are!
- New lazy `return_disk_surf_data` function added to DiskModelsPAGN.py
- New `readonly_cached_property` which is actually readonly
- New interpolator wrapping objects in `objects/interp.py` with better boundary support
- New AGNDiskInterp and AGNDisk model in `objects/disk.py`, using those interpolator wrappers, with tests
- Hoisted some things in `mcfacts.py` into function.

### Validation Steps
```bash
make install
pytest
make plot_scaling_setup
```

### Checklist

- [X] Commit messages follow the Conventional Commits guidelines
- [] Added / updated documentation


JakePost (approved):
Looks good! Local merge succeeds, all tests pass, and fiducial plots look as expected.

PR #389, opened 2026-09-03
Restore optimized variants and implement stars optimizations in restructure

nrposner:
We restore optimized function variants now that restructure has merged, including a new version of gw_strain_freq_helper that returns three arrays instead of two, following the changes to gw_strain_freq. That helper function requires mcfast version 0.1.10, which is now up on PyPI for all supported platforms.

In addition, we implement the optimizations for stars previously proposed in #375. That PR was closed, and is fully replaced by this one.

These changes lead to a ~2x speedup in restructure, from 148 seconds to 76 seconds testing the default run locally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ dependencies = [
     "pagn",
     "joblib>=1.4.2",
     "scikit_learn>=1.5.2",
-    "mcfast==0.1.8",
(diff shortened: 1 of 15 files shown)


JakePost (approved):
Looks good, everything runs well local and fiducial plots show the same results.

PR #390, opened 2026-09-04
Profiler Fix

JakePost:
Update implementation of cProfile and reorganize logic for handling the CLI subcommands.


nrposner (approved):
Ran locally, the resulting profile is substantively the same, and the logic makes more sense here. Approved.

PR #391, opened 2026-09-04
optimizing copy() and consistency_check()

nrposner:
Following PR #389 , the biggest single time sinks in restructure are copy() and consistency_check(). 

Jake observed that copy() deep-copies unnecessarily, since numpy arrays are immutable, so we can instead just use shallow copies, which dramatically reduces our memory use and time spent copying. This brings restructure time down from 76 seconds to 40 seconds.

consistency_check can be sped up considerably by taking a fast-path, just turning all the arrays into sets and making sure their mutual intersection is 0, or that the length of the merged sets is the same as the original arrays. Only in the case where we find a duplicate do we go through with list_occurrences() and actually find where the duplicates are for an informative error message. The fast path doesn't need that information. This brings restructure time down from 40 seconds to 33.

Total improvement is ~2.3x speedup with no difference to the output.

### DIFF
--- a/src/mcfacts/objects/agn_object_array.py
+++ b/src/mcfacts/objects/agn_object_array.py
@@ -786,16 +786,27 @@ class FilingCabinet:
 
 
     def consistency_check(self):
+        arrs = [(k, a.unique_id) for k, a in self.agn_objects.items() if k not in self.ignore_check]
+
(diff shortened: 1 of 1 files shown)


JakePost (approved):
Looks good!

PR #401, opened 2026-09-15
Fixed bug of separating SMBH accretion rate from (stellar mass) BH accretion rates 

varunprit:
### Summary of Changes
Changed disk_bh_eddington_ratio to be separated to smbh_eddinton_ratio and disk_bh_eddington_ratio in multiple instances. The list of files and functions left alone are as follows:
- most of scripts in recipes as they are for past papers
- mcfacts_sim.py (under the main function there are multiple times it calls the disk_bh_eddinton_ratio rate for BBH mergers)
- vera_mstar_bins.py
- in ReadInputs, settings_manager, mcfacts_default split the two up and added the smbh ratio to the top (defaulting to 0.5 per Barry's comment)
- in accretion.py functions change_bh_mass, change_bh_spin, change_bin_mass, change_bin_spin_magnitudes, change_bin_spin_angles, class ProgradeBlackHoleBondi(TimelineActor), class ProgradeBlackHoleAccretionTimelineActor),class BinaryBlackHoleAccretion(TimelineActor)
- dynamics.py
- migration.py

### Validation Steps
`make plots`

### Checklist

- [] Commit messages follow the Conventional Commits guidelines
- [] Added / updated documentation
- [] I have added the appropriate label to this Pull Request


src/mcfacts/modules/merge.py
```diff
@@ -827,21 +827,25 @@ def merged_orb_ecc(bin_orbs_a, v_kicks, smbh_mass, r_g_in_meters):
     ----------
     bin_orbs_a : numpy.ndarray
         Location of binary [r_{g,SMBH}] wrt to the SMBH with :obj:`float` type
-    v_kicks : numpy.ndarray
-        Kick velocity [km/s] with :obj:`float` type
+    bh_kick_comp_merged : numpy.ndarray
```
xevra: What's going on here?

xevra (commented):
CI is currently failing, and should be fixed.

The documentation for the new parameter looks good!

I don't have the time to take a second look at every instance of "disk_bh_eddington_ration" and double check if it is SMBH or stellar BH, and so I'm neither approving or rejecting this PR with my review.

I probably won't have time for it this week, so if somebody else could pick it up I would appreciate it.

JakePost:
I think these changes ~should~ need to be reauthorized in a fresh branch off of main-dev.

src/mcfacts/modules/merge.py
varunprit: Deleted it

varunprit:
Did not include in the comments but i found in disk.py there is some redundant setting of variables when they get redefined later to be the same name just lower case and spelled slighlty different (e.g. disk_alpha_viscosity = DISK_ALPHA_VISCOSITIES line 210) might be worth getting rid of those

